PR #3, opened 2025-03-04
Inject dynamic ATR version

netomi:
This PR injects a dynamic version based on the latest tag of the repo.

This works in both, a development environment when using `make serve` and when building a docker image using `make build`.

For that purpose, I added a dev dependency to dunamai that does the magic.

I have also created a PR for the infrastructure-p6 repo to properly configure the vm: https://github.com/apache/infrastructure-p6/pull/2098

Once that PR is merged, we can also merge that one and then the app running should have the proper version included.


netomi:
changing to long hash format is easy.
I did not find a way to get rid of the VERSION and COMMIT env variables.

Tried different approaches, but thats the cleanest things I could come up with.

sbp:
Removed the VERSION and COMMIT variables in [a949a7b0](https://github.com/apache/tooling-atr-experiments/commit/a949a7b094120da0ab8bef9c0c4435ded0c90bc4). Still need to do the long hash.

PR #76, opened 2025-05-07
Update tutorial.html minor text edits and questions

cottage14:
line 21 - If this is a tutorial, let's use the full name for the service at the start, and the TLA  for it later 
line 58 - avoid passive voice.
line 58 - 'chosen mailing list' presents a problem to the user of the tutorial. Who chooses it, how, and when in the process? line 67 - change 'report' to 'vote'

I note the use of 'use' in relation to clicking a button. Would not 'click' be clearer, unless there is some other behavior that is possible?

### DIFF
--- a/atr/templates/tutorial.html
+++ b/atr/templates/tutorial.html
@@ -18,7 +18,7 @@
 
 {% block content %}
   <h1>Tutorial for release managers</h1>
-  <p>The ATR is under development; the current version may not precisely match the screenshots below.</p>
+  <p>Adobe Trusted Releases (ATR) is under development; the current version may not precisely match the screenshots below.</p>
   <p class="btn-group my-3" role="group" aria-label="Tutorial sections">
     <a href="#compose" class="btn btn-outline-secondary">Compose</a>
     <a href="#vote" class="btn btn-outline-secondary">Vote</a>
@@ -55,7 +55,7 @@
          alt="Illustration: Review the results of your checks, and use the Start voting button." />
   </p>
   <p>
-    A form to start the vote is displayed. If necessary, use its ability to preview the vote announcement email body. When you submit the form, ATR sends an announcement message in your name to the chosen mailing list.
+    A form to start the vote appears. If necessary, use its ability to preview the vote announcement email body. When you submit the form, ATR sends an announcement message in your name to the chosen mailing list.
   </p>
   <p>
     <img src="{{ url_for('static', filename='png/tutorial-compose-04.png') }}"
@@ -64,7 +64,7 @@
 
   <h2 id="vote">Vote</h2>
   <p>
-    The <em>Vote</em> page is available to all project participants, and displays the report results.
+    The <em>Vote</em> page is available to all project participants, and displays the vote results.
   </p>
   <p>
     <img src="{{ url_for('static', filename='png/tutorial-vote-01.png') }}"


atr/templates/tutorial.html
```diff
@@ -18,7 +18,7 @@
 
 {% block content %}
   <h1>Tutorial for release managers</h1>
-  <p>The ATR is under development; the current version may not precisely match the screenshots below.</p>
+  <p>Adobe Trusted Releases (ATR) is under development; the current version may not precisely match the screenshots below.</p>
```
sbp, line 21: Should be `Apache` rather than `Adobe`.

atr/templates/tutorial.html
```diff
@@ -55,7 +55,7 @@ <h2 id="compose">Compose</h2>
          alt="Illustration: Review the results of your checks, and use the Start voting button." />
   </p>
   <p>
-    A form to start the vote is displayed. If necessary, use its ability to preview the vote announcement email body. When you submit the form, ATR sends an announcement message in your name to the chosen mailing list.
+    A form to start the vote appears. If necessary, use its ability to preview the vote announcement email body. When you submit the form, ATR sends an announcement message in your name to the chosen mailing list.
```
sbp, line 58: I originally wrote "appears", but it sounded strange to me and I struggled to find a replacement. I would certainly accept reverting it to "appears", but alternative suggestions to choose from would be welcome.

atr/templates/tutorial.html
```diff
@@ -64,7 +64,7 @@ <h2 id="compose">Compose</h2>
 
   <h2 id="vote">Vote</h2>
   <p>
-    The <em>Vote</em> page is available to all project participants, and displays the report results.
+    The <em>Vote</em> page is available to all project participants, and displays the vote results.
```
sbp, line 67: The _Vote_ page does not, perhaps counter-intuitively, display the vote results yet. As the tutorial itself says, "ATR does not yet tally votes; you must do this manually." We certainly plan this as a very important feature, but it will involve heuristics and a lot of testing so it is deferred for now. The reference to "report results" parallels the previous section: the reports are the same as those displayed to the user during the _Compose_ phase. When the _Compose_ phase is over, the reports are immutable during the _Vote_ phase. They are displayed to the release manager(s) during the _Compose_ phase to help inform them about what they need to change, fix, improve. They are then displayed during the _Vote_ phase again to help inform project participants what issues have been automatically detected. One thing we might consider is allowing the release manager to annotate the check reports so that e.g. participants are aware why errors have been manually, deliberately ignored.

sbp (changes requested):

sbp:
To answer a couple of other questions not covered in my PR review:

> Who chooses it, how, and when in the process?

The user chooses it using a radio form control when filling in the form to start the vote. But, in a wider sense, we choose it for the user as well, because we have strictly limited the available options for testing. This is the reason behind this admittedly unusual, vague, and passive wording.

Please note that in general I aimed for brutal concision in this tutorial. As documented internally (which is no help to you or other reviewers), we are thinking about much more detailed documentation and how to integrate that. The problem is that contemporary users are, I think, much more used to concision than detail. The problem with concision is that it does tend to leave out rather a lot of useful information.

> I note the use of 'use' in relation to clicking a button. Would not 'click' be clearer, unless there is some other behavior that is possible?

I prefer to [avoid talking about mechanics](https://www.w3.org/Provider/Style/NoMechanics), and I like to talk about navigation in a media independent way. I know that for the vast majority of cases a button is clicked, and that in any case this is more of a fossilised skeuomorphism, but this is how I think since being involved with the WAI in the early 2000s.

sbp:
I have attempted to address all of this feedback in 4a1532ec8b8d7850bda7e5cee7f14c1da0a5ad54.

sbp:
Assuming that 4a1532ec8b8d7850bda7e5cee7f14c1da0a5ad54 resolves all of the issues. If not, please reopen this PR, open an issue, or ping me on Slack as appropriate. Thanks very much for the feedback!

PR #251, opened 2025-10-20
Closes #243: Return more user friendly error messages throughout

### LINKED ISSUE #243: Return more user friendly error messages throughout
For example, permissions errors should always be 4xx not 5xx.

oujezdsky:
This PR addresses issue #243 by making error messages more user-friendly throughout the codebase.
I focused on rewriting vague or generic messages to be more explicit, actionable, and helpful, while also adjusting a few status codes to better reflect the nature of the errors.

No instances were found where 5xx codes were incorrectly used instead of 4xx (e.g., no 500 for permissions errors), so those weren't modified.


dave2wave:
Thank you for your contribution. Even though some linter checks failed, we'll clean those up soon.

PR #303, opened 2025-11-05
Expand pre-commit config

jbampton:
Add license header and fill out config and add two more "meta" hooks


jbampton:
I just saw the identity hook is actually a debugging hook.  Looks like we could remove this.

https://pre-commit.com/#meta-identity

dave2wave:
Thanks.

PR #309, opened 2025-11-10
Make the distribution forms more type safe

andrewmusselman:
Following established pattern to use Pydantic, moving from `forms.py` to `form.py`.

### DIFF
--- a/atr/form.py
+++ b/atr/form.py
@@ -20,6 +20,7 @@ from __future__ import annotations
 import enum
 import json
 import pathlib
+import re
 import types
 from typing import TYPE_CHECKING, Annotated, Any, Final, Literal, TypeAliasType, get_args, get_origin
 
@@ -43,6 +44,8 @@ if TYPE_CHECKING:
 
 DISCRIMINATOR_NAME: Final[str] = "variant"
 DISCRIMINATOR: Final[Any] = schema.discriminator(DISCRIMINATOR_NAME)
+
+_CONFIRM_PATTERN = re.compile(r"^[A-Za-z0-9 .,!?-]+$")
 
 
 class Form(schema.Form):
@@ -257,6 +260,7 @@ def render(  # noqa: C901
     border: bool = False,
     wider_widgets: bool = False,
     skip: list[str] | None = None,
+    confirm: str | None = None,
 ) -> htm.Element:
     if action is None:
         action = quart.request.path
@@ -320,7 +324,17 @@ def render(  # noqa: C901
         unused = ", ".join(custom.keys())
         raise ValueError(f"Custom widgets provided but not used: {unused}")
 
-    return htm.form(form_classes, action=action, method="post", enctype="multipart/form-data")[form_children]
+    form_attrs: dict[str, str] = {
+        "action": action,
+        "method": "post",
+        "enctype": "multipart/form-data",
+    }
+    if confirm:
+        if not _CONFIRM_PATTERN.match(confirm):
+            raise ValueError(f"Invalid characters in confirm message: {confirm!r}")
+        form_attrs["onsubmit"] = f"return confirm('{confirm}');"
+
+    return htm.form(form_classes, **form_attrs)[form_children]
 
 
 def render_block(block: htm.Block, *args, **kwargs) -> None:
@@ -731,7 +745,7 @@ def _render_field_value(
     return field_value
 
 
-def _render_row(
+def _render_row(  # noqa: C901
     field_info: pydantic.fields.FieldInfo,
     field_name: str,
     flash_error_data: dict[str, Any],
@@ -761,7 +775,10 @@ def _render_row(
     if widget_type == Widget.HIDDEN:
         attrs = {"type": "hidden", "name": field_name, "id": field_name}
         if field_value is not None:
-            attrs["value"] = str(field_value)
+            if isinstance(field_value, enum.Enum):
+                attrs["value"] = field_value.value
+            else:
+                attrs["value"] = str(field_value)
         return htpy.input(**attrs), None
 
     label_text = field_info.description or field_name.replace("_", " ").title()
--- a/atr/get/distribution.py
+++ b/atr/get/distribution.py
@@ -15,9 +15,10 @@
 # specific language governing permissions and limitations
 # under the License.
 
+
 import atr.blueprints.get as get
 import atr.db as db
-import atr.forms as forms
+import atr.form as form
 import atr.htm as htm
 import atr.models.sql as sql
 import atr.post as post
@@ -66,16 +67,6 @@ async def list_get(session: web.Committer, project: str, version: str) -> str:
     ## Distributions
     block.h2["Distributions"]
     for dist in distributions:
-        delete_form = await shared.distribution.DeleteForm.create_form(
-            data={
-                "release_name": dist.release_name,
-                "platform": dist.platform.name,
-                "owner_namespace": dist.owner_namespace,
-                "package": dist.package,
-                "version": dist.version,
-            }
-        )
-
         ### Platform package version
         block.h3(
             # Cannot use "#id" here, because the ID contains "."
@@ -94,13 +85,28 @@ async def list_get(session: web.Committer, project: str, version: str) -> str:
             shared.distribution.html_tr_a("Web URL", dist.web_url),
         ]
         block.table(".table.table-striped.table-bordered")[tbody]
-        form_action = util.as_url(post.distribution.delete, project=project, version=version)
-        delete_form_element = forms.render_simple(
-            delete_form,
-            action=form_action,
-            submit_classes="btn-danger",
+
+        delete_form = form.render(
+            model_cls=shared.distribution.DeleteForm,
+            action=util.as_url(post.distribution.delete, project=project, version=version),
+            form_classes=".d-inline-block.m-0",
+            submit_classes="btn-danger btn-sm",
+            submit_label="Delete",
+            empty=True,
+            defaults={
+                "release_name": dist.release_name,
+                "platform": shared.distribution.DistributionPlatform.from_sql(dist.platform),
+                "owner_namespace": dist.owner_namespace or "",
+                "package": dist.package,
+                "version": dist.version,
+            },
+            confirm=(
+                f"Are you sure you want to delete the distribution "
+                f"{dist.platform.name} {dist.package} {dist.version}? "
+                f"This cannot be undone."
+            ),
         )
-        block.append(htm.div(".mb-3")[delete_form_element])
+        block.append(htm.div(".mb-3")[delete_form])
 
     title = f"Distribution list for {project} {version}"
     return await template.blank(title, content=block.collect())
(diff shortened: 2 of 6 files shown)


atr/config.py
```diff
@@ -86,16 +86,7 @@ class AppConfig:
...
-            "cwells",
-            "dfoulks",
-            "fluxo",
-            "gmcdonald",
-            "humbedooh",
-            "sbp",
-            "tn",
-            "wave",
-        }
+        {"cwells", "dfoulks", "fluxo", "gmcdonald", "humbedooh", "sbp", "tn", "wave"}
```
sbp: Please revert unassociated changes to make the PR easier to review.

atr/get/distribution.py
```diff
@@ -66,55 +68,116 @@ async def list_get(session: web.Committer, project: str, version: str) -> str:
...
         block.table(".table.table-striped.table-bordered")[tbody]
-        form_action = util.as_url(post.distribution.delete, project=project, version=version)
-        delete_form_element = forms.render_simple(
-            delete_form,
-            action=form_action,
-            submit_classes="btn-danger",
-        )
-        block.append(htm.div(".mb-3")[delete_form_element])
+
+        # Create inline delete form with confirmation dialog (following projects.py pattern)
```
sbp: This is currently the correct way to do this, but please add a TODO to the effect that we should add a `confirm: bool` keyword argument to `form.render`. This will allow us to use `form.render` and not have to special case forms with confirmations.

atr/get/distribution.py
```diff
@@ -66,55 +68,116 @@ async def list_get(session: web.Committer, project: str, version: str) -> str:
...
+                htpy.i(".bi.bi-trash"), " Delete"
+            ],
+        ]
+        block.append(htm.div(".mb-3")[delete_form])
 
     title = f"Distribution list for {project} {version}"
     return await template.blank(title, content=block.collect())
 
 
+# The delete_get function can now be removed since we're using inline confirmation
```
sbp: Let's just remove this.

atr/get/distribution.py
```diff
@@ -66,55 +68,116 @@ async def list_get(session: web.Committer, project: str, version: str) -> str:
...
+
+
 @get.committer("/distribution/record/<project>/<version>")
 async def record(session: web.Committer, project: str, version: str) -> str:
-    form = await shared.distribution.DistributeForm.create_form(data={"package": project, "version": version})
-    fpv = shared.distribution.FormProjectVersion(form=form, project=project, version=version)
-    return await shared.distribution.record_form_page(fpv)
+    await shared.release_validated(project, version, staging=False)
+
+    block = htm.Block()
```
sbp: All of this code before `form.render` is duplicated in the `stage` function below, except for the call to `staging` in `html_nav_phase`. We can lift it into a helper function with `staging` parameterised.

atr/post/distribution.py
```diff
@@ -58,33 +57,14 @@ async def delete(session: web.Committer, project: str, version: str) -> web.Werk
...
-    match len(form.errors):
-        case 0:
-            await quart.flash("Ambiguous submission errors", category="warning")
-        case 1:
-            await quart.flash("There was 1 submission error", category="error")
-        case _ as n:
-            await quart.flash(f"There were {n} submission errors", category="error")
-    return await shared.distribution.record_form_page(fpv, staging=True)
+@post.form(shared.DistributeForm)
+async def stage_post(session: web.Committer, form: shared.DistributeForm, project: str, version: str) -> str:
```
sbp: We don't call forms `form` because that conflicts with the `atr.form` module name when imported in our standard style. For `ExampleForm` we use the name `example_form`, so for `DistributeForm` that would be `distribute_form`.

atr/storage/writers/distributions.py
```diff
@@ -161,8 +164,12 @@ async def record_from_data(
         match api_oc:
             case outcome.Result(result):
                 pass
-            case outcome.Error():
-                raise storage.AccessError("Failed to get API response from distribution platform")
+            case outcome.Error(error):
+                # Log the actual error for debugging
+                import atr.log as log
```
sbp: This should be top level unless it causes an import cycle.

atr/storage/writers/distributions.py
```diff
@@ -23,6 +23,7 @@
 
 import aiohttp
 import sqlalchemy.exc as exc
+from sqlalchemy import text
```
sbp: We don't use this import style, [according to our code conventions](https://release-test.apache.org/docs/code-conventions).

atr/get/distribution.py
```diff
@@ -66,55 +68,116 @@ async def list_get(session: web.Committer, project: str, version: str) -> str:
...
+    # Render the distribution form
+    form_html = form.render(
+        model_cls=shared.DistributeForm,
+        submit_label="Record distribution",
+        action=util.as_url(post.distribution.stage_post, project=project, version=version),
+        defaults={"package": project, "version": version},
+    )
+    block.append(form_html)
+
+    return await template.blank("Record Staging Distribution", content=block.collect())
```
sbp: We use sentence case for titles, [as mentioned in our code conventions](https://release-test.apache.org/docs/code-conventions).

atr/get/distribution.py
```diff
@@ -66,55 +68,116 @@ async def list_get(session: web.Committer, project: str, version: str) -> str:
...
+    # Render the distribution form
+    form_html = form.render(
+        model_cls=shared.DistributeForm,
+        submit_label="Record distribution",
+        action=util.as_url(post.distribution.record_post, project=project, version=version),
+        defaults={"package": project, "version": version},
+    )
+    block.append(form_html)
+
+    return await template.blank("Record Manual Distribution", content=block.collect())
```
sbp: We use sentence case for titles, [as mentioned in our code conventions](https://release-test.apache.org/docs/code-conventions).

atr/shared/distribution.py
```diff
@@ -36,55 +35,64 @@
...
+            "GitHub owner, ArtifactHub repo). Leave blank if not used."
+        ),
+        widget=form.Widget.TEXT,
+    )
+    package: str = form.label("Package", widget=form.Widget.TEXT)
+    version: str = form.label("Version", widget=form.Widget.TEXT)
+    details: bool = form.label(
+        description="Include details",
+        documentation="Include the details of the distribution in the response",
+        widget=form.Widget.CHECKBOX,
```
sbp: This can be dropped, like with `TEXT`.

atr/shared/distribution.py
```diff
@@ -36,55 +35,64 @@
...
+        documentation=(
+            "Who owns or names the package (Maven groupId, npm @scope, Docker namespace, "
+            "GitHub owner, ArtifactHub repo). Leave blank if not used."
+        ),
+        widget=form.Widget.TEXT,
+    )
+    package: str = form.label("Package", widget=form.Widget.TEXT)
+    version: str = form.label("Version", widget=form.Widget.TEXT)
+    details: bool = form.label(
+        description="Include details",
```
sbp: We typically don't use `description=` and `documentation=` keywords for the arguments to `form.label`.

atr/post/distribution.py
```diff
@@ -58,33 +57,14 @@ async def delete(session: web.Committer, project: str, version: str) -> web.Werk
...
-            # Should not happen
-            await quart.flash("Ambiguous submission errors", category="warning")
-        case 1:
-            await quart.flash("There was 1 submission error", category="error")
-        case _ as n:
-            await quart.flash(f"There were {n} submission errors", category="error")
-    return await shared.distribution.record_form_page(fpv)
+@post.form(shared.DistributeForm)
+async def record_post(session: web.Committer, form: shared.DistributeForm, project: str, version: str) -> str:
+    # Pydantic validation happens automatically in @post.form
```
sbp: We'd have to add this comment to every single POST handler to be consistent. Let's just remove this comment instead.

atr/post/distribution.py
```diff
@@ -58,33 +57,14 @@ async def delete(session: web.Committer, project: str, version: str) -> web.Werk
...
-        case 0:
-            await quart.flash("Ambiguous submission errors", category="warning")
-        case 1:
-            await quart.flash("There was 1 submission error", category="error")
-        case _ as n:
-            await quart.flash(f"There were {n} submission errors", category="error")
-    return await shared.distribution.record_form_page(fpv, staging=True)
+@post.form(shared.DistributeForm)
+async def stage_post(session: web.Committer, form: shared.DistributeForm, project: str, version: str) -> str:
+    # Pydantic validation happens automatically in @post.form
```
sbp: Let's remove this comment.

atr/shared/distribution.py
```diff
@@ -36,55 +35,64 @@
...
+
+        # Set default if needed and not provided
+        if default_owner_namespace and not self.owner_namespace:
+            self.owner_namespace = default_owner_namespace
+
+        # Validate requirements
+        if requires_owner_namespace and not self.owner_namespace:
+            raise ValueError(f'Platform "{self.platform.name}" requires an owner or namespace.')
+
+        if not requires_owner_namespace and not default_owner_namespace and self.owner_namespace:
```
sbp: Please ensure that all subexpressions are parenthesized, as pointed out in our [code conventions](https://release-test.apache.org/docs/code-conventions). Same goes for prior lines in this method, and potentially elsewhere.

atr/storage/writers/distributions.py
sbp: The [PEP8 convention for denoting private interfaces](https://peps.python.org/pep-0008/#public-and-internal-interfaces) is to use a single leading underscore in the interface name. We can use any interface of any dependency which does not begin with a single underscore.

We do not rename interfaces when we import them, as our code conventions state. Our standard form is `import a.b as b`, but if importing as `b` causes problems the best alternative is usually `import a.b`.

andrewmusselman:
Tested recording and deleting a new distribution after requested changes, works.

atr/form.py
```diff
@@ -69,6 +70,51 @@ class Widget(enum.Enum):
     URL = "url"
 
 
+class DistributionPlatformWrapper(enum.Enum):
```
andrewmusselman: @sbp just realized this enum and mapping should probably go in `shared/distribution` if you agree

atr/get/distribution.py
```diff
@@ -94,27 +86,80 @@ async def list_get(session: web.Committer, project: str, version: str) -> str:
...
         block.table(".table.table-striped.table-bordered")[tbody]
-        form_action = util.as_url(post.distribution.delete, project=project, version=version)
-        delete_form_element = forms.render_simple(
-            delete_form,
-            action=form_action,
-            submit_classes="btn-danger",
-        )
-        block.append(htm.div(".mb-3")[delete_form_element])
+
+        # TODO: add a confirm: bool keyword argument to form.render, allowing us to
```
sbp: This should be fairly trivial, so perhaps it would be a better idea to implement that within this PR? Otherwise there's a big block of code that we're adding here which will have to be entirely changed.

atr/post/distribution.py
```diff
@@ -17,7 +17,7 @@
 
 from __future__ import annotations
 
-import quart
+from typing import cast
```
sbp: Per the code conventions, we do not use `cast` or type ignores unless there is strictly no other alternative.

atr/post/distribution.py
```diff
@@ -28,17 +28,68 @@
 import atr.web as web
 
 
+async def record_form_process_page(
```
sbp: This interface is not alphabetically ordered.

atr/post/distribution.py
```diff
@@ -28,17 +28,68 @@
...
+                staging=staging,
+                dd=dd,
+            )
+        except storage.AccessError as e:
+            # Instead of calling record_form_page_new, redirect with error message
+            return await session.redirect(
+                get.distribution.stage if staging else get.distribution.record,
+                project=project,
+                version=version,
+                error=str(e),  # Flash error message
```
sbp: Per code conventions, we do not use line-ending comments.

atr/shared/distribution.py
```diff
@@ -17,74 +17,125 @@
...
+class DeleteForm(form.Form):
+    release_name: str = form.label("Release name", widget=form.Widget.HIDDEN)
+    platform: str = form.label("Platform", widget=form.Widget.HIDDEN)
+    owner_namespace: str = form.label("Owner namespace", widget=form.Widget.HIDDEN)
+    package: str = form.label("Package", widget=form.Widget.HIDDEN)
+    version: str = form.label("Version", widget=form.Widget.HIDDEN)
+
+
+class DistributeForm(form.Form):
+    platform: form.Enum[DistributionPlatformWrapper] = form.label(description="Platform")
```
sbp: We don't use keywords with `form.label`. I guess the wider convention is that we don't use keywords for arguments before `*`, but I'm not sure how widely that holds and if it's really a worthwhile convention.

atr/shared/distribution.py
```diff
@@ -17,74 +17,125 @@
...
+    version: str = form.label("Version", widget=form.Widget.TEXT)
+    details: form.Bool = form.label(
+        "Include details",
+        "Include the details of the distribution in the response",
+    )
+
+    @pydantic.field_validator("platform", mode="before")
+    @classmethod
+    def validate_platform(cls, value: Any) -> DistributionPlatformWrapper:
+        if isinstance(value, DistributionPlatformWrapper):
```
sbp: Which type does this method actually receive in practice?

atr/form.py
```diff
@@ -640,6 +640,10 @@ def _get_widget_type(field_info: pydantic.fields.FieldInfo) -> Widget:  # noqa:
     if annotation in (int, float):
         return Widget.NUMBER
 
+    # Check for plain enum types (including wrapper enums)
```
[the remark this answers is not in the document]
andrewmusselman: I was doing this to get the platform mapping working but with other fixes this isn't needed.

atr/post/distribution.py
```diff
@@ -28,17 +28,68 @@
...
+
 @post.committer("/distribution/delete/<project>/<version>")
-async def delete(session: web.Committer, project: str, version: str) -> web.WerkzeugResponse:
-    form = await shared.distribution.DeleteForm.create_form(data=await quart.request.form)
-    dd = distribution.DeleteData.model_validate(form.data)
+@post.form(shared.distribution.DeleteForm)
+async def delete(
+    session: web.Committer, delete_form: shared.distribution.DeleteForm, project: str, version: str
+) -> web.WerkzeugResponse:
+    dd = distribution.DeleteData.model_construct(**delete_form.model_dump())
```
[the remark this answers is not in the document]
andrewmusselman: Done

PR #336, opened 2025-11-20
Make the draft forms more type safe

andrewmusselman:
Tested making and deleting a draft, added a js validation on delete draft field and button since I don't see how that was done before.

### DIFF
--- a/atr/post/draft.py
+++ b/atr/post/draft.py
@@ -27,7 +27,7 @@ import quart
 
 import atr.blueprints.post as post
 import atr.construct as construct
-import atr.forms as forms
+import atr.form as form
 import atr.get as get
 import atr.log as log
 import atr.models.sql as sql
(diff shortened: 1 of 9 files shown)


atr/templates/check-selected.html
```diff
@@ -213,5 +184,21 @@ <h3 id="delete-draft" class="mt-4">Delete this draft</h3>
   {{ super() }}
   <script>
       init();
+
+      // Delete form validation
```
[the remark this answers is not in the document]
andrewmusselman: @sbp [a0851b0](https://github.com/apache/tooling-trusted-releases/pull/336/commits/a0851b06dd163bf0423440b4547e8ac822c3ccaf) fixes this; please review, thanks!

atr/post/draft.py
```diff
@@ -37,37 +36,12 @@
 import atr.web as web
 
 
-class VotePreviewForm(forms.Typed):
-    body = forms.textarea("Body")
-    # TODO: Validate the vote duration again?
```
sbp: This comment was lost in the new class. Is the comment, or any of its three lines, important? Should it have been retained, in part or in whole?

atr/shared/draft.py
```diff
@@ -15,21 +15,29 @@
...
-    file_path = forms.string("File path")
-    submit = forms.submit("Delete file")
+class DeleteFileForm(form.Form):
+    file_path: str = form.label(description="File path")
 
 
-class DeleteForm(forms.Typed):
-    """Form for deleting a candidate draft."""
+class DeleteForm(form.Form):
+    release_name: str = form.label(description="Release name", widget=form.Widget.HIDDEN)
```
sbp: The `description=` keyword should be removed from all of these `label` calls.

atr/shared/draft.py
```diff
@@ -15,21 +15,29 @@
...
+    @classmethod
+    def validate_confirm_delete(cls, value: str) -> str:
+        if value != "DELETE":
+            raise ValueError("You must enter 'DELETE' in this field")
+        return value
+
+
+class VotePreviewForm(form.Form):
+    body: str = form.label(description="Body", widget=form.Widget.TEXTAREA)
+    vote_duration: form.Int = form.label(description="Vote duration")
```
sbp: Should not use `description=`.

sbp (changes requested):
There's a problem with deleting a file: the HTML modal dialogue comes up with a new line of text saying `hidden_tag`, and submitting the form in the modal dialogue does not delete the file or even close the dialogue.

The HTML modal dialogue has always been buggy, but it does currently work on `main` for deleting files. We should take this opportunity to standardise on the new `confirm` mechanism that #309 introduces. Once #309 is squashed to a single commit we'll merge it, and then this PR can be updated to use it. Please also remove the entire HTML modal dialogue mechanism, as long as it's not used anywhere else.

sbp (changes requested):
Please also fix the Heisenberg merge errors, if there are any!

andrewmusselman:
I didn't touch those, will fix that.

PR #343, opened 2025-11-22
Update docs for WTForms -> Pydantic; fixes #338

### LINKED ISSUE #338: Update docs for WTForms -> Pydantic move
Updates needed:
* https://release-test.apache.org/docs/overview-of-the-code#user-interface
* https://release-test.apache.org/docs/user-interface#forms

andrewmusselman:
`make docs` run, reviewed layout locally


atr/docs/user-interface.md
```diff
@@ -115,43 +189,67 @@ The block class also adds a `data-src` attribute to elements, which records whic
...
+        # Process the validated form data...
+
+        await quart.flash("OpenPGP key added successfully.", "success")
+    except web.FlashError as e:
+        await quart.flash(str(e), "error")
+
+    return await session.redirect(get.keys.keys)
+```
 
-If you use the programmatic rendering functions from [`forms`](/ref/atr/forms.py), you can skip the template entirely. These functions return htpy elements, which you can combine with other htpy elements and return directly from the route, which is often useful for admin routes, for example. You can also use [`template.blank`](/ref/atr/template.py:blank), which renders a minimal template with just a title and content area. This is useful for simple pages that do not need the full template machinery.
```
andrewmusselman: @sbp I took this out because it had to do with the old `forms` module but if we want to keep any info let me know

atr/docs/user-interface.md
```diff
@@ -46,30 +46,104 @@ Template rendering happens in a thread pool to avoid blocking the async event lo
...
+    except web.FlashError as e:
+        await quart.flash(str(e), "error")
+    except Exception as e:
+        log.exception("Error adding OpenPGP key:")
+        await quart.flash(f"An unexpected error occurred: {e!s}", "error")
+
+    return await session.redirect(get.keys.keys)
+```
+
+The [`form.validate`](/ref/atr/form.py:validate) function takes the form class, the form data dictionary, and an optional context dictionary. If validation succeeds, it returns an instance of your form class with validated data. If validation fails, it raises a `pydantic.ValidationError`.
```
sbp: This is true, but as noted above, `form.validate` is not used in the sample code and should be avoided except in certain circumstances.

PR #357, opened 2025-11-26
Add markdownlint with pre-commit

### LINKED ISSUE #347: Add Markdown linting
Our [tooling-docs](https://github.com/apache/tooling-docs/blob/main/.github/linters/.markdown-lint.yml) have Markdown linting, but we do not presently have any such lints in ATR. We should add some.

jbampton:
Add basic markdownlint config file

closes #347 

Rules are from here:

https://github.com/DavidAnson/markdownlint?tab=readme-ov-file#rules--aliases

cmark generate HTML

### DIFF
--- a/.github/linters/.markdown-lint.yml
+++ b/.github/linters/.markdown-lint.yml
@@ -1,0 +1,17 @@
+MD004:
+  style: consistent
+
+MD013:
+  code_blocks: false
+  line_length: 915
+  tables: false
+
+MD033: false
+
+MD034: false
+
+MD040: false
+
+MD041: false
+
+MD046: false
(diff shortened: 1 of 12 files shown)


sbp:
Would it be possible to use the list in [this comment](https://github.com/apache/tooling-trusted-releases/issues/347#issuecomment-3572261836) please?

Also the `.html` pages are just fragments used in other templates. Perhaps it would be more correct if we used some other file name extension to indicate that? What do you think? But the `<html>` etc. wrappers would break all the pages that they're used in.

Thanks for the work so far!

sbp:
@jbampton Could you tell us what version of `cmark` you have, please?

sbp:
I'm planning to move HTML generation into the container. We've been discussing this recently, and I think this is more evidence that we should do that. I like having the HTML files locally for easy browsing, but I can still build and export them to e.g. a directory in `.gitignore`.

sbp:
Done in ee5a3bfe442ac9317ca0f427e335945ab397ebfd. All the HTML files have gone away. You might want to rebase this PR onto that commit!

PR #370, opened 2025-11-29
README: Remove minor underline from workflow buttons

jbampton:
I tested in both Chrome and Opera the minor blue marks appear.

Just thought I would bring this up.  Please close this PR if not needed. Thanks !

## Old look

atr-buttons-blue-underline

## New look

Screenshot from 2025-11-30 06-42-18

So the cleaner buttons are also seen here:

https://github.com/jbampton/tooling-trusted-releases/blob/86c88dc53e380cf6e586c361355e3936b2946d92/README.md


sbp:
Thanks!

Issue #364, opened 2025-11-28
Record when committees are automatically updated


sbp:
We automatically sync with remote data, but it may be useful for us to track what happened and when.

sbp:
This is a successor issue to #277.

developer-ravi-03:
Hi @sbp 👋  
I recently contributed PR #449 and would like to work on this issue next.

My plan is to record automatic committee updates in the audit log, including timestamp and source of update.

Please let me know if this approach sounds good before I start implementation.
Thanks!

asf-tooling:
**Automated triage** — analyzed at `main@2da7807a`

**Type:** `new_feature`  •  **Classification:** `actionable`  •  **Confidence:** `medium`
**Application domain(s):** `project_committee_management`, `shared_infrastructure`

### Summary
The issue requests recording what happens when committees are automatically synced from remote data sources (Whimsy, projects.apache.org). @sbp noted this can be recorded in the audit log. @developer-ravi-03 volunteered to implement it via PR but no response was given. The `update_metadata()` function in `atr/datasources/apache.py` already returns aggregate counts but doesn't record specific changes (which committees were added/modified, what fields changed). The enhancement would add detailed change tracking to the metadata sync process.

### Where this lives in the code today

#### `atr/datasources/apache.py` — `update_metadata` (lines 298-334)
_needs modification_
This is the main entry point for metadata syncing. It returns only aggregate counts but doesn't record details about what specifically changed.

```python
async def update_metadata() -> tuple[int, int]:
    """Update metadata from remote data sources."""

    ldap_projects = await get_ldap_projects_data()
    projects = await get_projects_data()
    podlings_data = await get_current_podlings_data()
    committees = await get_active_committee_data()

    ldap_projects_by_name: Mapping[str, LDAPProject] = {p.name: p for p in ldap_projects.projects}
    committees_by_name: Mapping[str, Committee] = {c.name: c for c in committees.committees}

    added_count = 0
    updated_count = 0

    async with db.session() as data:
        async with data.begin():
            added, updated = await _update_committees(data, ldap_projects, committees_by_name)
            added_count += added
            updated_count += updated

            added, updated = await _update_podlings(data, podlings_data, ldap_projects_by_name)
            added_count += added
            updated_count += updated

            added, updated = await _update_projects(data, projects)
            added_count += added
            updated_count += updated

            added, updated = await _update_tooling(data)
            added_count += added
            updated_count += updated

[...]

PR #374, opened 2025-11-30
Record timestamp when committees are automatically updated

Abhay-raj-srivastava:
This PR implements the request from Issue #364: "Record when committees are automatically updated."

# What this change does
- Adds timestamp recording inside `update_metadata()` after `_update_committees()` runs.
- Stores the value in metadata using:
  `last_auto_committee_update`
- Uses `datetime.now(timezone.utc).isoformat()` for consistent UTC formatting.

# Why this is needed
The tooling updates committees automatically, but until now the system did not keep track of when this update occurred. This timestamp allows better observability and debugging for automated sync events.

# Summary
- Added metadata timestamp write
- Maintained existing structure and async session
- Ensured the update is recorded only after a successful committee sync

Thank you! Please let me know if any changes are required.

### DIFF
--- a/atr/datasources/apache.py
+++ b/atr/datasources/apache.py
@@ -34,6 +34,8 @@ import atr.models.helpers as helpers
 import atr.models.schema as schema
 import atr.models.sql as sql
 import atr.util as util
+from datetime import datetime, timezone
+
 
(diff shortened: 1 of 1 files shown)


sbp:
@Abhay-raj-srivastava Thank you for your contribution. We only accept PRs from new contributors after they have introduced themselves on our mailing list, as documented in the README of this project. In addition, the code of this PR causes the update task to crash.

PR #377, opened 2025-12-01
Add top nav bar

dave2wave:
Needs more work and review.


sbp:
@dave2wave Should be fixed in 812bd3614a30efd4bf4eef0a94cb21eb5de83f37.

sbp (commented):
**Current status:**

We intend to replace the sidebar with this design. We will probably run the two designs concurrently for a while because tests will need to be migrated from one to the other. I plan also to port the templates to `atr.html`. @dave2wave plans to specify which items will go into which menu options.

dave2wave:
I tried rebasing and got a bunch of changes that really should not be in the branch.

1. uv.lock
2. All of the png files that were compressed.

PR #418, opened 2025-12-11
SBOM report updates

alitheg:
- Added support for including the source component in the conformance error model.
- Skipped missing component errors for "file" components since they don't have PURLs etc.
- Fixed some wording and added a hyperlink to vulnerability details.


atr/sbom/conformance.py
```diff
@@ -289,7 +289,8 @@ def ntia_2021_issues(
             cpe_is_none = bom_value.metadata.component.cpe is None
             purl_is_none = bom_value.metadata.component.purl is None
             swid_is_none = bom_value.metadata.component.swid is None
-            if cpe_is_none and purl_is_none and swid_is_none:
+            type_is_file = bom_value.metadata.component.type == "file"
+            if cpe_is_none and purl_is_none and swid_is_none and not type_is_file:
```
sbp: Needs parentheses around `not type_is_file`.

alitheg: Thanks, I'm just reading the rest of the style guide now :)

PR #422, opened 2025-12-15
Docker build improvements

alitheg:
- Split docker build into 3 stages
- Simplify dependencies
- Use uv for python in ubuntu since we use it anyway
- Use buildkit for parallel builds
- Use docker cache during build

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -14,7 +14,7 @@
 .pytype/
 .ruff_cache/
 .tox/
-.venv*
+.venv
 __pycache__
 cert.pem
 dmypy.json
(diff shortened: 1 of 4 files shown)


sbp:
In a simplified benchmark, i.e. using whatever caching was already to hand, this reduces the build time of the Alpine container from 3m18.879s to 2m2.920s. Using the prior `RUN curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /usr/local/bin` approach to installing syft, though less secure in the long run because it doesn't pin, reduces the build time further to just 0m47.321s. I'll change this in a subsequent commit. Presumably the script is downloading binaries; perhaps there's a way that we can download them but used pinned versions? Maybe the script itself already supports that somehow?

The `--no-cache` flag was added in 3cffe561607cf40143cb0b6a8497d8484ac30b72 and I recall that I did so only for debugging. I left it in because I hadn't been building the container locally very often. Now that we have local reload in containers for development, it should still be relatively uncommon, but probably doesn't matter too much either way.

PR #430, opened 2025-12-19
Update outdated tool scanners and add ATR tool metadata to the SBOM

alitheg:
SBOM-modifying operations now log their tool usage in the SBOM tools section. The report will pick up on this and note any changes. I've also refactored the outdated tools check to support multiple types of tool - for now just the existing CycloneDX maven plugin and ATR itself, but this is now expandable for future use.

The CDX spec defines an incrementing version field which should be updated when anything modifies the BOM. I've added this update in and attached the produced/scanned version as result data on the tasks. Finally, I've unified this with the task of pulling together and patching the SBOM where relevant, if all modifications go through that method then we'll know it always increments correctly.

### DIFF
--- a/atr/sbom/tool.py
+++ b/atr/sbom/tool.py
@@ -1,0 +1,104 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from __future__ import annotations
+
+import datetime
+from typing import TYPE_CHECKING, Any, Final
+
+import semver
+
+from . import maven, models, utilities
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+
+_KNOWN_TOOLS: Final[dict[str, models.tool.Tool]] = {
+    # name in file: ( canonical name, friendly name, version callable )
+    "cyclonedx-maven-plugin": models.tool.Tool("cyclonedx-maven-plugin", "CycloneDX Maven Plugin", maven.version_as_of),
+    "cyclonedx maven plugin": models.tool.Tool("cyclonedx-maven-plugin", "CycloneDX Maven Plugin", maven.version_as_of),
+    "apache trusted releases": models.tool.Tool(
+        "apache trusted releases", "Apache Trusted Releases platform", lambda _: utilities.get_atr_version()
+    ),
+}
+
+
+def plugin_outdated_version(bom_value: models.bom.Bom) -> list[models.tool.Outdated] | None:
+    if bom_value.metadata is None:
+        return [models.tool.OutdatedMissingMetadata()]
+    timestamp = bom_value.metadata.timestamp
+    if timestamp is None:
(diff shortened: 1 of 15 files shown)


atr/sbom/utilities.py
```diff
@@ -164,3 +195,35 @@ def _map_severity(severity: str) -> str:
         if sev == "moderate":
             return "medium"
     return "unknown"
+
+
+def _record_task(task: str, revision: str, doc: yyjson.Document, patch_ops: models.patch.Patch) -> models.patch.Patch:
+    properties: list[dict[str, str]] | None = get_pointer(doc, "/properties")
```
sbp: Are all of these types definitely over `dict[str, str]`?

atr/sbom/tool.py
```diff
@@ -0,0 +1,108 @@
...
+
+try:
+    from atr import metadata
+
+    atr_version = metadata.version
+except ImportError:
+    metadata = None
+    atr_version = "cli"
+
+_KNOWN_TOOLS: Final[dict[str, tuple[str, str, Callable[[str], str | None]]]] = {
```
sbp: Maybe better to use a Pydantic model, dataclass, or namedtuple etc. for the values here.

atr/sbom/tool.py
```diff
@@ -0,0 +1,108 @@
...
+
+import datetime
+from typing import TYPE_CHECKING, Any, Final
+
+from . import maven, models
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+try:
```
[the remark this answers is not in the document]
alitheg: Yes I was wondering about this. I didn't fix it as any fix needs to live in the `sbom` folder to preserve the ability to relative import. But I could add a method to `utilities.py` to return the version (and then perhaps we could add a CLI version as well)

atr/sbom/utilities.py
sbp: Even tools? Because I don't understand:

```
    tools: dict[str, str] | None = get_pointer(doc, "/metadata/tools")
    services: list[dict[str, str]] | None = get_pointer(doc, "/metadata/tools/services")
```

If services are in tools, and tools is `{str: str}` but service values are themselves `{str: str}`, doesn't this mean that some of the values of tools are `{str: str}`?

atr/tasks/sbom.py
```diff
@@ -145,31 +144,29 @@ async def osv_scan(args: FileArgs) -> results.Results | None:
...
-                new_full_path = os.path.join(str(creating.interim_path), args.file_path)
-                # Write to the new revision
-                log.info(f"Writing updated SBOM to {new_full_path}")
-                await aiofiles.os.remove(new_full_path)
-                async with aiofiles.open(new_full_path, "w", encoding="utf-8") as f:
-                    await f.write(merged.dumps())
-
-            if creating.new is None:
-                raise RuntimeError("Internal error: New revision not found")
+    new_version, merged = sbom.utilities.apply_patch("osv-scan", args.revision_number, bundle, patch_ops)
```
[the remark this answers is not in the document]
sbp: Oh yeah, actually that does make sense entirely. "No vulnerabilities found" is of course a valid outcome worth recording in the SBOM! Let's leave it like this.

atr/sbom/utilities.py
sbp: Rather, that some of the values of tools are `[{str: str}]`, since services is a list. But in any case, not `str`.

alitheg: Yes, you're right - sorry I saw the comment was on properties and didn't look any further!

atr/sbom/tool.py
```diff
@@ -0,0 +1,108 @@
...
+        return None
+    # If the version used is less than the version available
+    if version_comparable[0] < expected_version_comparable[0]:
+        # Then note the version available
+        return expected_version
+    # Otherwise, the user is using the latest version
+    return None
+
+
+def version_parse(version_str: str) -> tuple[tuple[int, int, int], str] | None:
```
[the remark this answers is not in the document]
alitheg: Agreed, that would make sense. And the conventions in there follow semver (in the way the current ATR version doesn't - but can deal with that). I'll move this to a semver parser and refuse to check if it doesn't match.

PR #442, opened 2025-12-23
Pull previous SBOM results into the report and highlight new/changed vulnerabilities/licenses

alitheg:
- Includes previous vulns and licenses in SBOMToolScore task results
- Previous calculated using lexographic version sort
- Shows changes to each on the report
- Reorganises report get function 
- License scanner now optionally returns category A licenses in addition to B and X
- Add icon function to htm for bootstrap icons

### DIFF
--- a/atr/get/sbom.py
+++ b/atr/get/sbom.py
@@ -18,7 +18,7 @@
 from __future__ import annotations
 
 import json
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 
 import asfquart.base as base
 import cmarkgfm
@@ -27,6 +27,8 @@ import markupsafe
 import atr.blueprints.get as get
 import atr.db as db
 import atr.form as form
+import atr.get.compose as compose
+import atr.get.vote as vote
 import atr.htm as htm
 import atr.models.results as results
 import atr.models.sql as sql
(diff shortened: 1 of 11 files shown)


alitheg:
@sbp Feel free to leave comments/thoughts/suggestions for tomorrow - or to merge if you think it's fine.

atr/get/sbom.py
```diff
@@ -468,58 +547,86 @@ def _vulnerability_scan_find_in_progress_task(
...
-                        _cdx_to_osv(v)
-                        for v in vulns
-                        if (v.affects is not None) and (component in [a.get("ref") for a in v.affects])
-                    ],
-                ),
-            )
+    if ignored_count > 0:
+        component_word = "component was" if (ignored_count == 1) else "components were"
+        block.p[f"{ignored_count} {component_word} ignored due to missing PURL or version information:"]
+        block.p[f"{','.join(ignored)}"]
```
sbp: I think this is missing a `block.append(new_block)` after it, to make it work the same as `_vulnerability_results_from_bom` which does do that.

atr/get/sbom.py
```diff
@@ -378,45 +388,96 @@ def _missing_tally(items: list[sbom.models.conformance.Missing]) -> list[tuple[s
...
+            })"
         if key not in components:
-            components[key] = [str(item)]
+            components[key] = [name]
         else:
-            components[key].append(str(item))
+            components[key].append(name)
     return sorted(
-        [(category, count, components.get(category, [])) for category, count in counts.items()],
+        [(category, count, new_count, components.get(category, [])) for category, count in counts.items()],
```
sbp: The `new_count` is the count of overall new licenses, by the looks, but these returned rows are per category. Shouldn't the new count therefore be tracked per category rather than overall, and be custom for each category row?

atr/get/sbom.py
```diff
@@ -183,41 +182,48 @@ def _outdated_tool_section(block: htm.Block, task_result: results.SBOMToolScore)
 
 
 def _conformance_section(block: htm.Block, task_result: results.SBOMToolScore) -> None:
+    block.h2["Conformance report"]
     warnings = [sbom.models.conformance.MissingAdapter.validate_python(json.loads(w)) for w in task_result.warnings]
     errors = [sbom.models.conformance.MissingAdapter.validate_python(json.loads(e)) for e in task_result.errors]
     if warnings:
-        block.h2["Warnings"]
+        block.h3[htm.icon("bi-exclamation-triangle-fill", ".me-2.text-warning"), "Warnings"]
```
sbp: The `htm.icon` function adds the `bi-` prefix, so this will have `bi-bi-` as a prefix.

atr/sbom/models/base.py
```diff
@@ -21,7 +21,7 @@
 
 
 class Lax(pydantic.BaseModel):
-    model_config = pydantic.ConfigDict(extra="allow", strict=False)
+    model_config = pydantic.ConfigDict(extra="allow", strict=False, validate_assignment=True, validate_by_name=True)
```
sbp: I wonder what happens if we add `validate_assignment` to `schema.Lax`? It's probably not a good idea for us to let `schema.py` and `base.py` get out of sync. Similarly, if `validate_by_name` is needed for SBOM related parsing, then we should probably add it to `schema.Lax` too.

atr/get/sbom.py
```diff
@@ -56,16 +59,55 @@ async def report(session: web.Committer, project: str, version: str, file_path:
...
+
+    return await template.blank("SBOM report", content=block.collect())
+
+
+async def _fetch_tasks(
+    file_path: str, project: str, release: sql.Release, version: str
+) -> tuple[sql.Task | None, collections.abc.Sequence[sql.Task], collections.abc.Sequence[sql.Task]]:
     async with db.session() as data:
         via = sql.validate_instrumented_attribute
-        # TODO: Abstract this code and the sbomtool.MissingAdapter validators
```
sbp: Do we not still want to do this?

atr/get/sbom.py
```diff
@@ -378,45 +388,96 @@ def _missing_tally(items: list[sbom.models.conformance.Missing]) -> list[tuple[s
...
     for item in items:
         key = item.category
         counts[key] = counts.get(key, 0) + 1
+        name = str(item).capitalize()
+        if item.component_name not in old_map:
+            new_count = new_count + 1
+            name = f"{name} (new)"
+        elif item.license_expression != old_map[item.component_name][0]:
+            new_count = new_count + 1
+            name = f"{name} (previously {old_map[item.component_name][0]} - Category {
```
sbp: Could make this (and "(new)" above) `<strong>`. Maybe a diff style + OLD - NEW change over two lines would be better, or `OLD -> new` or something like that. Some colour might help too.

atr/get/sbom.py
alitheg: Maybe the validators - I assumed part of the TODO was pulling them out of the main route method. I'll re-add

atr/get/sbom.py
alitheg: It's slightly odd (and could use some improvement I'm sure) - but this function is actually called once for each category, so although this would return the overall count if you did pass licenses with multiple categories - because warnings are all category B and errors are all category X, it does render the correct information

atr/get/sbom.py
```diff
@@ -378,45 +388,96 @@ def _missing_tally(items: list[sbom.models.conformance.Missing]) -> list[tuple[s
...
         if vuln.references is not None:
             vuln_refs = [r for r in vuln.references if r.get("type", "") == "WEB"]
         vuln_primary_ref = vuln_refs[0] if (len(vuln_refs) > 0) else {}
         vuln_modified = vuln.modified or "Unknown"
+
         vuln_severity = _extract_vulnerability_severity(vuln)
+        worst = _update_worst_severity(severities, vuln_severity, worst)
+
+        if previous_vulns is not None:
+            if (vuln_id not in previous_vulns) or previous_vulns[vuln_id] != vuln_severity:
```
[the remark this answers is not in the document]
alitheg: I think what I've done does this now - I updated the list of previous vulnerabilities to include the affected bom-ref from the CDX version. The OSV schema doesn't include the component because it's just the vulnerability itself.

Issue #282, opened 2025-11-03
Warn the user if the "Preserve download files" option is checked during announcing


sbp:
This will make file movement errors less mysterious. We should also give the same error when such movement errors occur.

Supersedes #278.

developer-ravi-03:
Hi @sbp / @dave2wave 👋  
I’d like to work on this issue.  
Please let me know if I can proceed or if you have any additional guidance.  
Thanks!

sbp:
Thanks @developer-ravi-03. Please go ahead. We already have a warning box on the announce form that says "**TODO**: The limited options above are provided for testing purposes. In the finished version of ATR, you will be able to send to your own specified mailing lists." I think that the style of this would be useful to resolve the present issue too. It could be placed just below the documentation (on the right hand side of the two columns, in other words), in the "Download path suffix" field, and could say something like "WARNING: The preserve download files option is enabled, which means that if any existing download paths match new paths in this release, then the release will fail."

developer-ravi-03:
Thanks @sbp! That’s very helpful.
I’ll proceed with adding the warning in the location you suggested and follow the existing style.

PR #474, opened 2026-01-05
Warn user when preserve download files is enabled during announce

developer-ravi-03:
Adds a warning message below the download path suffix field when
"Preserve download files" is enabled, as discussed in issue #282.


sbp:
Thanks for catching the CI error. I've [asked the team to allow biomejs](https://github.com/apache/infrastructure-actions/pull/440).

Issue #477, opened 2026-01-05
Document identification of generated source files in lightweight and RAT license checks


dave2wave:
User documentation should share the file path patterns used to exclude license header checks on generated files.

sanjana2505006:
Hello @dave2wave,
Just wanted to check whether this issue is still open to work on.
I’d be happy to take it up if so.

dave2wave:
This documentation can only be completed properly after #427 has been completed for lightweight checks. #427 is partially completed for RAT checks.

sanjana2505006:
@dave2wave,
Thank you for the clarification. 
That makes sense, I understand that the documentation here depends on the completion of #427, especially for the lightweight checks.

Please let me know whether you’d prefer that I:
	•	keep the PR open and update it once #427 is fully completed, or
	•	Pause/close it for now and revisit this issue after #427 is done.

Happy to align with whatever workflow you think works best.

Issue #427, opened 2025-12-18
Automatically exclude `.min.js` and other generated files from RAT and License Header checks


sbp:
These files are considered compiled, in a sense, and do not need to contain licenses.

sbp:
Resolved by 40013cbb7dcb72c5557289ef54974f9159da02dc.

Issue #274, opened 2025-10-30
New compose project options for license check excludes 


dave2wave:
Add new text arrays as compose project options:

- Source excludes patterns.
- Binary excludes patterns.

These align with _Source artifact paths_ and _Binary artifact paths_. The release manager can paste these into project options from whatever their source is. If these are empty then fallback to looking for an included _rat excludes file_.

Editing these while composing a release would invalidate prior checks and cause a new evaluation. When running the checks a temporary excludes file can be created.

Let's discuss an action button on the compose page.

PR #481, opened 2026-01-06
fixed issue 477

sanjana2505006:
## Issue:
Users were not aware of how the system identifies generated files in order to exclude them from mandatory licence header checks. Although this logic existed in the code, it was not documented. As a result, users had difficulty understanding why certain files were skipped or how to configure their own exclusions.

## Changes I made:
I analyzed the source code for both the Apache RAT check and the lightweight licence check to determine the exact rules being applied. I then documented these rules in the user guide so users can easily reference and understand the exclusion behaviour.

This PR fixes the issue #477


sanjana2505006:
Hello @dave2wave,
I’ve opened an issue and linked the related PR. Whenever you have time, I’d appreciate a review. Thank you for your consideration!

sbp:
@sanjana2505006 Issues #274 and #427 are now both complete, so continued work on this PR should be unblocked.

sanjana2505006:
Hello! I've addressed the feedback:

1. **Index Update**: I've added `license-checks.md` to `atr/docs/index.md` and ran the build script to sync the navigation.
2. **Squashed & Rebased**: I've squashed all changes into a single clean commit relative to the current `upstream/main`.
3. **Verified Checks**: I've verified that all lint checks (`make check`) pass perfectly on the finalized branch.

sbp:
@sanjana2505006 Thanks for sticking with this. This most recent commit seems to have duplicated much of the content into [`atr/docs/user-guide.md`](https://github.com/apache/tooling-trusted-releases/pull/481/files#diff-1d0f75a439ea3326da17770b2f4e91bdeaa41b9025ecedc3e378fd1ac5e5ffd6). Only `atr/docs/index.md` needed to be changed (and was changed, correctly).

PR #492, opened 2026-01-09
New year

sebbASF:


NOTICE
```diff
@@ -1,9 +1,9 @@
 Apache Trusted Releases (ATR)
 
-Copyright 2025 The Apache Software Foundation
+Copyright 2026 The Apache Software Foundation
 
 This product includes software developed at
-The Apache Software Foundation (/).
+The Apache Software Foundation.
```
sbp: The [NOTICE file section](https://www.apache.org/legal/src-headers.html#notice) of the _ASF Source Header and Copyright Notice Policy_ says to use `The Apache Software Foundation (/).`

sbp (commented):
What should we do about `(/)`?

PR #501, opened 2026-01-12
Work to enable automatic distributions

alitheg:


atr/db/interaction.py
```diff
@@ -395,6 +388,36 @@ async def tasks_ongoing_revision(
...
 
 
+async def trusted_jwt(publisher: str, jwt: str, phase: TrustedProjectPhase) -> tuple[dict[str, Any], str, sql.Project]:
+    if publisher != "github":
+        raise InteractionError(f"Publisher {publisher} not supported")
+    payload = await jwtoken.verify_github_oidc(jwt)
+    if payload["actor_id"] != _GITHUB_TRUSTED_ROLE_NID:
+        asf_uid = await ldap.github_to_apache(payload["actor_id"])
+    else:
+        asf_uid = "arm"  # TODO: What to do here?
```
alitheg: @sbp Going to amend this commit when this is fixed - but what should we do here? See above where this JWT now comes from the tooling role instead of a named individual.

alitheg: Actually I think I've sorted this. Don't need the UID in this case.

sbp:
@alitheg Usually APIs accept different authentication mechanisms at each individual endpoint, so we should probably do that instead when functionality overlaps, as in the case of `/distribution/record`, `/release/announce`, and `/vote/resolve`. It might be a little difficult because we get the ASF UID and project from different sources in each case.

sbp:
@alitheg Yeah, exactly. Tracking that as #504 since this PR is about to be merged.

Issue #517, opened 2026-01-14
Document database cascades


sbp:
We allow some deletions to cascade in the database, but we have not documented what these cascades are. To understand them it is necessary to read the source, but it would be useful to explain them in the developer documentation.

dave2wave:
@sbp If I have analyzed this properly this about documenting the class relationships in `models/sql.py` and determining how deletion is handled for the `1-M` and `1-1` relationships: Does the `-C->` annotation indicate a Cascade?

1. Explicit deletions. Before the owner is deleted all of the subs are expressly deleted.
2. Cascade deletion. If the owner is deleted then the sub is automatically deleted by sqlalchemy.
3. Blocked deletion. If certain subs exist then deletion is blocked. When this happens an archive happens.

PR #519, opened 2026-01-14
Document database cascade deletions

Aditi102005:
This PR documents the database cascade for Release → CheckResult.
Deleting a Release automatically deletes all its CheckResult entries. 
This behavior is implemented using a CASCADE foreign key in migration 0008_2025.06.12_26c0022b.py.
References issue #517.

### DIFF
--- a/notes/database-cascades.md
+++ b/notes/database-cascades.md
@@ -1,0 +1,8 @@
+# Database Cascades
+Certain deletions in the database automatically cascade to related entities. This document lists them for clarity.
+
+| Parent Entity | Cascades To   | Migration File | Notes |
+|---------------|--------------|----------------|-------|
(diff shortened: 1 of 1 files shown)


sbp:
@Aditi102005 Thank you very much for this contribution! I was hoping that, instead, the documentation could be added to the databases documentation that we have in our existing developer documentation in `atr/docs/`. I was also expecting the new documentation to give an overview of how cascading works in SQLite and in the ORM (just briefly), including the direction of flow of deletions, and then to enumerate the full list of cascade linking in our existing models. This is quite a substantial amount of documentation and I expect it may not be a good first issue for contributors, which is why I didn't label it as such. You're welcome to continue working on it, and I'm grateful for the contribution so far, but I cannot accept this PR and, as I say, I think this would be quite a bit of work.

PR #529, opened 2026-01-15
Possible recurrent jobs implementation

alitheg:
Add the ability to schedule future jobs. Worker claim updated to not claim future tasks. Metadata and workflow status tasks can optionally schedule their next iteration.

### DIFF
--- a/atr/worker.py
+++ b/atr/worker.py
@@ -120,7 +120,8 @@ async def _task_next_claim() -> tuple[int, str, list[str] | dict[str, Any], str]
                     sqlmodel.and_(
                         sql.Task.status == task.QUEUED,
                         sqlmodel.or_(
-                            via(sql.Task.scheduled).is_(None), sql.Task.scheduled <= datetime.datetime.now(datetime.UTC)
+                            via(sql.Task.scheduled).is_(None),
+                            via(sql.Task.scheduled) <= datetime.datetime.now(datetime.UTC),
                         ),
                     )
                 )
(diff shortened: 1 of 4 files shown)


alitheg:
@sbp I realised we already have sufficient plumbing for scheduling, since the worker runs on a loop and claims tasks. 

There are some potential improvements I can think of to this - such as adding a "scheduled" column to Task instead of using "added" - but I didn't want to spend too long on it as a first pass. Let me know what you think of the concept and we can refine or throw it away!

atr/server.py
```diff
@@ -463,20 +464,36 @@ async def _initialise_test_environment() -> None:
...
-        await asyncio.sleep(86400)
+async def _register_recurrent_tasks() -> None:
+    """Schedule recurring tasks"""
+    # Wait one minute to allow the server to start
+    await asyncio.sleep(30)
+    try:
+        await tasks.clear_scheduled()
+        metadata = await tasks.metadata_update(asf_uid="system", schedule_next=True)
+        log.info(f"Scheduled remote metadata update with ID {metadata.id}")
+        workflow = await tasks.workflow_update(asf_uid="system", schedule_next=True)
```
sbp: Must prevent thundering herd here too.

atr/worker.py
```diff
@@ -114,7 +114,12 @@ async def _task_next_claim() -> tuple[int, str, list[str] | dict[str, Any]] | No
             # Get the ID of the oldest queued task
             oldest_queued_task = (
                 sqlmodel.select(sql.Task.id)
-                .where(sql.Task.status == task.QUEUED)
+                .where(
+                    sqlmodel.and_(
+                        sql.Task.status == task.QUEUED,
+                        sql.Task.added <= datetime.datetime.now(datetime.UTC) - datetime.timedelta(seconds=2),
```
sbp: `(datetime.datetime.now(datetime.UTC) - datetime.timedelta(seconds=2))`

sbp (changes requested):
This is a fantastic idea, thanks!

Issue #475, opened 2026-01-05
Allow files to be tagged in the compose phase according to a mapping


dave2wave:
There should be a compose policy field which is a yaml of tag to gitignore pattern matches.This mapping should be a default which tags files in certain subdirectories for distribution channels.

Something like:

```
maven:
  - 'maven/**'
pypi:
  - 'pypi/**'
dockerhub:
  - `docker/**'
```

hboutemy:
need to add ASF distribution = what is to be published to https://downloads.apache.org/ independently from third-party repositorie slike Maven Central, PyPI, Dockerhub, etc...

hboutemy:
just found https://github.com/apache/tooling-docs/blob/main/apache-trusted-releases/svn-dist.md = what i called previously "ASF distribution"

I'd expect that "ATR writes to SVN - ATR Releases write to svn:dist/release as an interim step" is done before we work on "ATR writes to third party distributions" https://github.com/apache/tooling-docs/blob/main/apache-trusted-releases/distributions.md

@dave2wave can you confirm that's the plan?

dave2wave:
You should rsync the files into the root directory so that on the GH Runner they all are at the root level. I was looking for a rich setup, but if it is easier then we can define the subdirectories that must be used.

PR #579, opened 2026-01-22
Allow projects to specify a tagging spec, allow rsync of tag

alitheg:
## Pull request summary 

**Allow projects to specify a tagging spec, allow rsync of tag**

**Description:**

This PR adds a tagging spec using glob patterns as specified in #475. It's stored in release policy, though I imagine this may change as part of the attestation work - or at least where we get it from at distribution time might change. 

In addition, I've updated rsync to support requesting a tag as well as a project/version - it will evaluate the glob patterns and modify the rsync command to send the matching files, flattened into the target directory.

---

## Required acknowledgements

Please replace each `[ ]` with `[x]` to confirm.
PRs missing confirmations may be closed or converted to Draft.

* [x] I have read and followed **CONTRIBUTING.md**
* [x] I have read **DEVELOPMENT.md**
* [x] I have run the required tests and checks locally
* [x] All required checks are currently passing
* [x] This branch is **rebased on the current `main` branch**

### DIFF
--- a/atr/ssh.py
+++ b/atr/ssh.py
@@ -455,6 +455,8 @@ async def _step_06a_validate_read_permissions(
         tags = release.release_policy.file_tag_mappings.keys()
         if tag not in tags:
             raise RsyncArgsError(f"Tag '{tag}' is not allowed for release '{release.name}'")
+        if ".." in "".join(release.release_policy.file_tag_mappings[tag]):
+            raise RsyncArgsError(f"Tag '{tag}' is misconfigured for release '{release.name}'")
         return release, release.release_policy.file_tag_mappings[tag]
     return release, None
 


atr/ssh.py
```diff
@@ -352,11 +403,15 @@ def _step_05_command_path_validate(path: str) -> tuple[str, str]:
...
         raise RsyncArgsError("The path argument should be a /PROJECT/VERSION/ directory path")
 
-    path_project, path_version = path.strip("/").split("/", 1)
+    path_project, path_version, *rest = path.strip("/").split("/", 2)
+    tag = rest[0] if rest else None
     alphanum = set(string.ascii_letters + string.digits + "-")
     if not all(c in alphanum for c in path_project):
         raise RsyncArgsError("The project name should contain only alphanumeric characters or hyphens")
 
+    if tag and (not all(c in alphanum for c in path_project)):
```
sbp: Presumably should be `tag` rather than `path_project` at the end.

migrations/versions/0041_2026.01.21_44cdc6b9.py
```diff
@@ -0,0 +1,27 @@
...
+# Revision identifiers, used by Alembic
+revision: str = "0041_2026.01.21_44cdc6b9"
+down_revision: str | None = "0040_2026.01.15_31d91cc5"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    with op.batch_alter_table("releasepolicy", schema=None) as batch_op:
+        batch_op.add_column(sa.Column("atr_file_tagging_spec", sa.JSON(), nullable=False, server_default="{}"))
```
sbp: How about `file_tag_mappings`? (I've applied this migration too, so you won't be the only one downgrading!)

atr/storage/writers/policy.py
```diff
@@ -95,13 +97,15 @@ async def edit_compose(self, form: shared.projects.ComposePolicyForm) -> None:
         project_name = form.project_name
         _, release_policy = await self.__get_or_create_policy(project_name)
 
+        atr_tags_dict = yaml.safe_load(form.atr_file_tagging_spec) or {}
```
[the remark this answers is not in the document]
sbp: Let's also check for path traversal here, as early as possible.

sbp:
> Thinking about it - this rsync approach does have one major drawback, that it always flattens the filesystem for the caller. This may be a problem - what do you think @sbp?

We definitely shouldn't flatten, but I think we should strip leading directories at least. So if the files are `abc/def.txt` and `abc/pqr/xyz.txt` we could strip the leading `abc/` at least. Hopefully that's a good compromise. In other words, the file closest to the root sets the base. There can of course be multiple bases if, e.g., we have `a/b.txt` and `p/q.txt`, and then we can't strip anything.

Issue #633, opened 2026-02-06
Login/logoff should be a button on the top nav and not a menu


dave2wave:
Login / Logoff are awkward and should be a button on the top nav and not a dropdown menu.

Abhishekmishra2808:
@dave2wave Hi! I’d like to work on this and prepare a PR. Could you please assign this issue to me?

Abhishekmishra2808:
Hi @dave2wave
 I raised a PR to address this issue! Please let me know if something is missing at my end.

PR #634, opened 2026-02-06
Move auth buttons

Abhishekmishra2808:
### Description

This change addresses issue #633 by improving the navigation UX. Previously, the Login/Logout action was placed inside a dropdown menu, which required extra clicks and felt unintuitive.

The update moves Login/Logout to a direct button in the top navigation bar. The button is rendered conditionally based on authentication state - showing **Log in** when the user is logged out and **Log out** when logged in - while preserving existing authentication logic and the account dropdown.

This results in faster access, clearer navigation behavior, and better overall usability.

**AFTER**
image
image

**TEST RESULTS**

image

### DIFF
--- a/atr/templates/includes/topnav.html
+++ b/atr/templates/includes/topnav.html
@@ -283,7 +283,7 @@
       <ul class="navbar-nav align-items-center">
         {% if current_user %}
           <li class="nav-item dropdown">
-            <button class="dropdown-toggle btn btn-sm btn-secondary ms-2 mt-1"
+            <button class="nav-link dropdown-toggle btn"
                     id="dropdownUser"
                     type="button"
                     data-bs-toggle="dropdown"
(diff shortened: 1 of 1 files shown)


dave2wave:
The buttons need to be lowered by about 3px so that the text aligns.

I like the Account dropdown, but please name it "User" and put it before the "Log Out" button.

Abhishekmishra2808:
Thanks! I’ll adjust the vertical alignment, rename the dropdown to “User,” and reorder it before the logout button. I’ll push an update shortly.

Abhishekmishra2808:
heya @dave2wave, I updated the PR! please look into it when you have some free time :)

dave2wave:
Closer
1. The `Log In` button is properly aligned!
2. The `Log Out`button is now too low. Change the style to match `Log In`
3. The `User` dropdown button should be styled like the other drop downs. Copy the style from `About The ASF`

Abhishekmishra2808:
I’ve implemented all the suggestions shared by @dave2wave. I was thinking of moving the `login `and `logout `buttons to the **far right**, since the logout button looked slightly misaligned in the middle to me. That was just my opinion - if you prefer your approach, I’ve updated the PR accordingly.

Issue #224, opened 2025-09-05
Adjustments for Alpha2


dave2wave:
There are adjustments for Alpha2:

1. Vote emails should go to the PMC's mailing lists. Announce emails continue to be redirected to the test mailing list. (Thinking about it)
2. Vote email template needs to have the template adjusted to include a note that it is from the Alpha Test.
3. They site warning message will be adjusted.

sbp:
Implemented parts 1 and 2 in 4555e568448a6905585d4111aba19f1f2d2832b4. All that remains is to modify the wording of the site warning message and link it to the new _About_ page, which has already been added.

Issue #201, opened 2025-07-23
Tie ATR server into the rsync path of svn dist artifacts


dave2wave:
This means implementation of transition 2 from https://github.com/apache/tooling-docs/blob/main/apache-trusted-release/svn-dist.md

If after discussion with Infra we could consider starting with transition 1.

andrewmusselman:
@dave2wave Was there an outcome from talking with infra?

dave2wave:
This issue depends on #19 and #245

dave2wave:
In parallel #284

dave2wave:
We are going to need a test svn dist repository for writes.

sbp:
There are some relevant notes about this here:

https://github.com/apache/tooling-trusted-releases/issues/753#issuecomment-3954808393

asf-tooling:
**Automated triage** — analyzed at `main@2da7807a`

**Type:** `discussion`  •  **Classification:** `no_action`  •  **Confidence:** `medium`
**Application domain(s):** `artifact_upload`, `cryptographic_keys`, `announcement_publishing`

### Summary
This is an ongoing architectural planning issue about integrating ATR into the rsync distribution path for SVN dist artifacts. The approach has evolved through discussion from 'transition 2' to '2b' to the latest decision by @dave2wave (2026-02-15) to use 'transition 1b'. The concrete work involves making the ATR container access a sparse checkout of the dist release SVN server and using it as DOWNLOADS_STORAGE_DIR. Work is tracked across multiple sub-issues (#19, #245, #649, #753), with @sbp most recently (16 days ago) pointing to relevant notes in #753.

### Where this lives in the code today

#### `atr/svn/commits.py` — `handle` (lines 34-39)
_currently does this_
Handles PubSub SVN commit notifications and updates a local working copy - part of the mechanism to keep ATR's view of SVN dist in sync.

```python
async def handle(payload: dict, working_copy_root: pathlib.Path) -> None:
    pubsub_path = str(payload.get("pubsub_path", ""))
    # Ignore commits outside dist/dev or dist/release
    if pubsub_path.startswith(_WATCHED_PREFIXES):
        log.debug(f"PubSub payload: {payload}")
        await _process_payload(payload, working_copy_root)
```

#### `atr/svn/__init__.py` — `commit` (lines 94-115)
_currently does this_
SVN commit capability - ATR can already write to SVN, which would be used in the transition to push release artifacts to dist.

```python
[...]

PR #649, opened 2026-02-14
clarify: svn:dist not done by ATR yet

hboutemy:
fixes message from #224

@alitheg @dave2wave AFAIK, ATR currently cannot svn commit to svn:dist/release, isn't it?

= "ATR writes to SVN Dist" step 1 from plan at https://tooling.apache.org/svn-dist.html not yet implemented

### DIFF
--- a/atr/templates/layouts/base.html
+++ b/atr/templates/layouts/base.html
@@ -48,7 +48,7 @@
           <main class="main-content">
             {% if (not current_user) or (current_user.uid != "sbp") %}
               <div class="warning-banner">
-                <strong>Alpha 2 Software:</strong> ATR is available as Alpha software. You can test the release process. Actual releases can be downloaded and committed to svn.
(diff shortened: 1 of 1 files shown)


hboutemy:
notice that currently ATR proposes to "Create a distribution
Create a distribution of maven-apache-resources-1.7 using the form below."

that proposes a lot of "platforms" = third party distributions channels ("Artifact Hub", "Docker hub", "Maven Central", "npm", "npm (scoped)" and "PyPI")

but it does not tall anything about Apache source distribution = what has been staged in previous steps

there is a huge confusion on this "distribution" between Apache source distribution (to download.apache.org) vs all the future third party binary distribution channel, that are not yet in the scope of ATR

dave2wave:
#201 is the issue that will allow distribution to happen. The question has been when.

I find that this red box at the top of every page to be very distracting and really don't want to expand it with detail. I propose that we go ahead and work on #201 soon. I f we want to have instructions on how to use ATR and then manually continue then the details should go onto the finish page.

dave2wave (changes requested):

dave2wave:
> I have only one concern for now on ATR staging vs svn:dist: I'm not convinced I can keep the same directory structure

Request a session with @sbp and he will show you how you should be able to accomplish what you need in the finish phase.

Issue #677, opened 2026-02-17
SSH server: Configure explicit cipher suites, KEX, and MAC algorithms


andrewmusselman:
**ASVS References:** 11.3.1 (INFO-02), 11.3.2 (Finding 1 — Medium)

### Description

The SSH server created via `asyncssh.create_server()` in `atr/ssh.py` (line ~147) does not specify explicit `encryption_algs`, `kex_algs`, or `mac_algs`. While asyncssh's current defaults are secure, relying on implicit defaults means:

- A client could potentially negotiate a legacy algorithm if one is supported by the library.
- Future library updates could change defaults without the application noticing.
- There is no auditable record of which algorithms are intentionally permitted.

### Current code

```python
server = await asyncssh.create_server(
    SSHServer,
    server_host_keys=[key_path],
    process_factory=process_factory,
    host=_CONFIG.SSH_HOST,
    port=_CONFIG.SSH_PORT,
    encoding=None,
)
```

### Recommended remediation

```python
APPROVED_CIPHERS = [
    'aes256-gcm@openssh.com',
    'aes128-gcm@openssh.com',
    'aes256-ctr',
    'aes128-ctr',
]

APPROVED_KEX = [
    'curve25519-sha256',
    'ecdh-sha2-nistp256',
    'diffie-hellman-group16-sha512',
]

APPROVED_MACS = [
    'hmac-sha2-256-etm@openssh.com',
    'hmac-sha2-512-etm@openssh.com',
]

server = await asyncssh.create_server(
    SSHServer,
    server_host_keys=[key_path],
    process_factory=process_factory,
    host=_CONFIG.SSH_HOST,
    port=_CONFIG.SSH_PORT,
    encoding=None,
    encryption_algs=APPROVED_CIPHERS,
    kex_algs=APPROVED_KEX,
    mac_algs=APPROVED_MACS,
)
```

### Severity

Medium — This is the only non-informational finding across all three audits. It is a defense-in-depth gap, not an active vulnerability.

sbp:
Resolved by 030b4fc39add34172696cc0c222fa4751db38d00.

PR #689, opened 2026-02-17
#677 - Add explicit ciphers, kex and mac algorithms.

alitheg:
I've created a PR for this one as I wanted to validate the choices. I tried to follow the defaults (which are slightly more broad than the original suggestion in #677) because that way we maintain the same compatibility we have today, however for the key exchange algorithms, the `asyncssh` module checks to see which capabilities are available and changes some of those defaults dynamically. We could do that checking ourselves but that feels unnecessary to me unless we specifically want to allow some of those options - hence please let me know what if any you'd like to add or remove from these lists!

### DIFF
--- a/atr/ssh.py
+++ b/atr/ssh.py
@@ -43,6 +43,36 @@ import atr.util as util
 
 _CONFIG: Final = config.get()
 
+_APPROVED_CIPHERS: Final[list[str]] = [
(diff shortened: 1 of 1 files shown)


alitheg:
Sean has implemented a better version of this in #690

PR #694, opened 2026-02-17
clarify ASF distribution vs third party channels

hboutemy:
@alitheg after our discussion from Friday: need to clarify ASF  (mainly source release) distribution = svn:dist/release / downloads.apache.org
vs third party channels = https://github.com/apache/tooling-docs/blob/main/apache-trusted-releases/distributions.md

the tooling-docs page itself require clarification, too: I'll propose a PR later


hboutemy:
ok, it seems I definitively don't understand the plan on ASF distribution directory vs third party channels

yes, I plan to have a discussion with @sbp on ASF distribution directory

but it definitively is independent from the topic about third party distribution channels

perhaps I'll better understand once I have the discussion done

PR #801, opened 2026-03-01
Add has post access check for controls for committers

### LINKED ISSUE #802: Allow released files to be read more widely
At least to anybody who is logged in. In alpha these files are never deleted and are the only way to browse what has been "released" in ATR. In beta these files will be deleted after 90-180 days, and are for showing the state when released only. Therefore in alpha it may be best to make them public, and in beta committer only, since they're only intended for debugging and we don't want people using them in CI or linking to them from their download pages.

dave2wave:
Closes #802 

We need to ensure that all committers can access many pages.

Let's collaborate on this PR.

### DIFF
--- a/atr/get/file.py
+++ b/atr/get/file.py
@@ -46,8 +46,6 @@ async def selected(
     URL: /file/<project_name>/<version_name>
     View all the files in a release (any phase).
     """
-    # has_post = await session.has_post_access(project_name)
-    # open page w/o a form
     release = await session.release(str(project_name), str(version_name), phase=None)
 
     revision_number = release.latest_revision_number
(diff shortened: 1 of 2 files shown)


alitheg:
I could be missing something but I think this would actually mean we don't need `check_access` for GET at _all_? Because we're already checking you're logged in by using `web.Committer` instead of `web.Public`. This way we could remove `check_access` from the `get.typed` decorator and all the GET routes would open up to any logged in user.

alitheg:
True - so we could call it from the get routes where required but just remove the (new) call to it on every GET route that includes `<project_name>`

PR #850, opened 2026-03-10
Drop `file_name` field in upload files

### LINKED ISSUE #780: Upload file path validation bypass when file_name parameter is provided
**Source:** ASVS 2.2.2 Finding 12
**CWE:** CWE-22 (Improper Limitation of a Pathname to a Restricted Directory)
**Location:** `atr/storage/writers/release.py`, lines 295–310

#### Description

The `upload_files` method has different code paths for filename validation. When the `file_name` parameter is provided, it bypasses `form.to_relpath()` validation:

```python
if not file_name:
    validated_path = form.to_relpath(file.filename)  # Validated
else:
    relative_file_path = file_name  # NOT validated
```

#### Recommendation

- Always validate the path through `form.to_relpath()` regardless of source.
- We should add a SafeFilename string type

dave2wave:
closes #780

### DIFF
--- a/atr/post/upload.py
+++ b/atr/post/upload.py
@@ -194,13 +194,16 @@ async def _add_files(
     version_name: safe.VersionName,
 ) -> web.WerkzeugResponse:
     try:
-        file_name = add_form.file_name
+        # file_name = add_form.file_name
         file_data = add_form.file_data
 
         async with storage.write(session) as write:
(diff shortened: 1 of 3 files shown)


sbp:
@dave2wave It looks like this field was actually being used. Shall I remove the code that implemented it, or would you like further review and discussion?

Issue #776, opened 2026-02-25
Vote content fields lack length and content validation


andrewmusselman:
**Source:** ASVS 2.2.1 Finding MEDIUM-003
**CWE:** CWE-20 (Improper Input Validation)
**Locations:**
- `atr/shared/vote.py` — `CastVoteForm`
- `atr/post/voting.py`, lines 95–105

#### Description

Vote comments, email bodies, and resolution text are accepted without length limits. Oversized content could cause email delivery failures, storage exhaustion, or abuse of mailing list archives.

#### Recommendation

- Add `max_length` constraints to vote comment fields (e.g., 5,000 characters)
- Add length limits to email body generation (e.g., 50,000 characters)

PR #894, opened 2026-03-16
#776 - implement 50k character limit for vote comment and support passing form errors through a cache in the form module instead of via flash in the session

alitheg:
It's worth noting this solution doesn't scale beyond a single server (or at least it would require sticky sessions).

### DIFF
--- a/atr/get/ignores.py
+++ b/atr/get/ignores.py
@@ -47,28 +47,29 @@ async def ignores(
     content = htm.div[
         htm.h1["Ignored checks"],
         htm.p[f"Manage ignored checks for project {project_key!s}."],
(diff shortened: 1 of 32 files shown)


atr/get/ignores.py
```diff
@@ -47,28 +47,29 @@ async def ignores(
     content = htm.div[
         htm.h1["Ignored checks"],
         htm.p[f"Manage ignored checks for project {project_key!s}."],
-        _add_ignore(str(project_key)),
-        _existing_ignores(ignores),
+        _add_ignore(session.asf_uid, str(project_key)),
```
sbp: `async` but not `await`ed.

Issue #901, opened 2026-03-16
Expand SBOM support


dave2wave:
Currently SBOM support is for CycloneDX JSON with a particular suffix. We need to be more general and should consider the following:

1. Add support for CycloneDX XML.
2. Add support for classifiers on possible suffixes.

I was going to add support for SPDX - I'm told Nuttx is working on this, but I'm not seeing anything in their releases and nothing specific on their mailing lists.

dave2wave:
@alitheg Sorry to drop this thread. Should we have a review and then decide if this is complete for now?

alitheg:
Yes, I was going to book a review with you guys. Will drop a message to the team.

PR #942, opened 2026-03-25
Not for merging (yet) #901 - add support for XML in sbom tooling

alitheg:


atr/sbom/models/bom.py
```diff
@@ -1,100 +0,0 @@
-# Licensed to the Apache Software Foundation (ASF) under one
```
sbp: This file was deleted, but `__init__.py` still imports it:

```python
from . import base, bom, bundle, conformance, licenses, osv, patch, sbomqs, tool

__all__ = ["base", "bom", "bundle", "conformance", "licenses", "osv", "patch", "sbomqs", "tool"]
```

atr/sbom/utilities.py
```diff
@@ -138,10 +140,35 @@ def patch_to_data(patch_ops: models.patch.Patch) -> list[dict[str, Any]]:
...
+    if path.name.endswith(".json"):
+        bom_json: dict[str, Any] = json.loads(text)
+        bom = Bom.from_json(data=bom_json)
+        source_type = "json"
+        version_str = bom_json.get('specVersion', '1.7')
+        spec_version = SchemaVersion.from_version(version_str)
+    elif path.name.endswith(".xml"):
+        bom_xml = ElementTree.fromstring(text)
+        bom = Bom.from_xml(bom_xml)
+        tag = re.match(r"\{http://cyclonedx.org/schema/bom/(.+)}", bom_xml.tag)
```
sbp: Should probably have an `if tag is None` guard after this for a better error message.

atr/sbom/osv.py
```diff
@@ -165,29 +166,6 @@ def _assemble_vulnerabilities(doc: yyjson.Document, patch_ops: models.patch.Patc
     )
 
 
-def _component_purl_with_version(component: models.bom.Component) -> str | None:
```
sbp: Why was this deleted? I thought that this was quite useful. Were we finding that there was a problem with it?

sbp (commented):
XML -> Bom object -> JSON -> Document object -> patched Document object -> JSON -> BOM -> XML is a bit of a long transformation process, and I'd worry that it's a bit fragile. I would rather try to convince projects to standardise on JSON SBOMs: the most recent survey I could find showed that JSON is used far more often than XML for SBOMs.

Issue #1003, opened 2026-03-31
SSH Interface Lacks Rate Limiting for Write Operations


asf-tooling:
**ASVS Level(s):** [L2-only]

**Description:**

### Summary
The SSH rsync interface has no rate limiting on write operations, while the web interface has comprehensive rate limiting. This creates a bypass path where authenticated users can perform unlimited writes via SSH while being throttled on the web interface, enabling release object creation flooding, parallel upload flooding, and resource exhaustion.

### Details
**Affected Files and Lines:**
- `atr/ssh.py` - SSH interface without rate limiting

The SSH interface allows unlimited write operations, creating an alternative path to bypass web interface rate limits. This inconsistency undermines the rate limiting security controls.

### Recommended Remediation
Implement SSH-specific rate limiting:

```python
# Track operations per ASF UID
_ssh_rate_limits: dict[str, collections.deque] = {}

def _check_ssh_rate_limit(asf_uid: str) -> bool:
    """Check if user has exceeded SSH rate limits."""
    now = time.time()
    
    if asf_uid not in _ssh_rate_limits:
        _ssh_rate_limits[asf_uid] = collections.deque()
    
    operations = _ssh_rate_limits[asf_uid]
    
    # Remove operations outside 1-hour window
    while operations and now - operations[0] > 3600:
        operations.popleft()
    
    # Check limits: 10 writes/minute, 100 writes/hour
    recent_minute = sum(1 for t in operations if now - t < 60)
    if recent_minute >= 10:
        return False
    if len(operations) >= 100:
        return False
    
    operations.append(now)
    return True
```

Add `_check_ssh_rate_limit()` function and call it in `_step_02_handle_safely()` before processing write operations. Implement periodic cleanup task for rate limit tracking data. Make timeout configurable via `atr/config.py` with `SSH_RSYNC_TIMEOUT` parameter.

### Acceptance Criteria
- [ ] Rate limiting implemented for SSH writes
- [ ] 10 writes/minute limit enforced
- [ ] 100 writes/hour limit enforced
- [ ] Rate limit tracking per user
- [ ] Cleanup task prevents memory growth
- [ ] Unit test verifying the fix

### References
- Source reports: L2:2.3.2.md
- Related findings: FINDING-027
- ASVS sections: 2.3.2

### Priority
High

---

PR #1147, opened 2026-04-09
#1003 - add rate limiting to SSH connections

alitheg:
Rate limiting based on IP and username, same as the web side. We IP rate limit in connection start, with a higher limit (to prevent automated attacks but still allow multiple legitimate users from one IP). In auth complete we rate limit per-user - so that we know the username was validated with a key (to prevent malicious usage of another user's rate limit)

### DIFF
--- a/atr/ssh.py
+++ b/atr/ssh.py
@@ -19,6 +19,7 @@
 
 import asyncio
 import asyncio.subprocess
+import collections
 import datetime
 import glob
 import os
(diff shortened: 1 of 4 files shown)


atr/ssh.py
```diff
@@ -251,6 +269,25 @@ def _output_stderr(process: asyncssh.SSHServerProcess, message: str) -> None:
...
 
 
+def _rate_limit_check(bucket: dict[str, collections.deque[float]], key: str, limit: int) -> bool:
+    """Return True if key is within the rate limit, False if exceeded. Mutates bucket."""
+    now = time.monotonic()
+    window = bucket.get(key)
+    if window is not None:
+        while window and (now - window[0]) > _RATE_WINDOW:
+            window.popleft()
+        if not window:
```
sbp: Ah, this doesn't work either. The problem is that when this is deleted we set `window = None` and then fall through. Then, later on, when `window is None`, we add a window to the bucket again. So really this clearing is a noop! But also, it doesn't make sense conceptually to drop the entry when we have a new request coming in. We can only drop all the IPs that aren't connecting, but that would mean scanning them all.

So, how about adding this to the cleanup task that you just added to `server.py`?

sbp (changes requested):

sbp (approved):

PR #1148, opened 2026-04-09
Use the debug print format that is used throughout osv.py

### LINKED ISSUE #1130: Debug print() Bypasses Structured Logging
**ASVS Level(s):** [L2-only]

**Description:**

### Summary
The file `atr/sbom/osv.py` at line 110 uses `print(vulns)` for debug output, bypassing the application's structured logging framework. This produces unstructured output that cannot be filtered, routed, or managed through the logging configuration, reducing log manageability and potentially exposing information through uncontrolled output channels.

### Details
The use of `print()` statements in production code:
- Bypasses logging level controls (cannot be disabled in production)
- Outputs to stdout instead of configured log handlers
- Produces unstructured text instead of structured log entries
- Cannot be filtered or routed by logging infrastructure
- May expose information in contexts where stdout is captured differently than logs

**Affected file:** `atr/sbom/osv.py`, line 110

This inconsistency with the application's structured logging approach reduces operational visibility and control.

### Recommended Remediation
Replace the `print()` statement with structured logging:

```python
log.debug("Loaded vulnerabilities from bundle", count=len(vulns))
```

If detailed vulnerability information is needed for debugging, use:

```python
log.debug("Loaded vulnerabilities from bundle", count=len(vulns), vulns=vulns)
```

### Acceptance Criteria
- [ ] Replace `print(vulns)` with structured logging call
- [ ] Verify log output includes vulnerability count
- [ ] Confirm logging respects configured log levels
- [ ] Search codebase for other `print()` statements and remediate

### References
- Source reports: L2:14.2.4.md
- Related findings: None
- ASVS sections: 14.2.4

### Priority
Low

---

dave2wave:
One way to close #1130

### DIFF
--- a/atr/sbom/osv.py
+++ b/atr/sbom/osv.py
@@ -123,7 +123,7 @@ def vulns_from_bundle(bundle: models.bundle.Bundle) -> list[models.osv.CdxVulner
     vulns = get_pointer(bundle.doc, "/vulnerabilities")
     if vulns is None:
         return []
-    print(vulns)
+    print(f"[DEBUG] {len(vulns)} vulnerabilities:\n{vulns}")
     return [models.osv.CdxVulnerabilityDetail.model_validate(v) for v in vulns]
 
 


alitheg:
Yes I think this is good, and we should add some sort of commentary to the audit guidance to say that the sbom module is used as a standalone CLI so doesn't hook into the standard structured logging.

Side thought - can/should we instantiate logs with some matching names in `cli.py` so that we *could* use the same log format as elsewhere?

alitheg (approved):

alitheg:
> > Side thought - can/should we instantiate logs with some matching names in cli.py so that we could use the same log format as elsewhere?
> 
> Ideally yes, but we have to be pragmatic. If you want to create a `future` issue then please go ahead.

Agreed this is the pragmatic solution for now. I've created #1149 to track it

PR #1150, opened 2026-04-09
Store session data in the server

sbp:
See also https://github.com/apache/infrastructure-asfquart/tree/sbp-server-side-sessions


atr/blueprints/common.py
```diff
@@ -54,24 +56,31 @@
...
 async def authenticate() -> web.Committer:
-    web_session = await asfquart.session.read()
-    if web_session is None:
+    web_session = await sessions.read()
+    if not isinstance(web_session, sql.UserSession):
         raise base.ASFQuartException("Not authenticated", errorcode=401)
-    if (web_session.uid is None) or (not await ldap.is_active(web_session.uid)):
-        asfquart.session.clear()
+
+    if not await ldap.is_active(web_session.uid):
```
alitheg: Do we want to go to LDAP every time we validate a session in addition to the periodic check in `before_request`? I assume this will be fine with the pubsub implementation (so we have a cache rather than a live check?)

atr/post/user.py
```diff
@@ -119,17 +119,16 @@ async def _cache_session(session: web.Committer) -> None:
...
-        "projects": getattr(session, "projects", []),
-        "mfa": getattr(session, "mfa", False),
-        "roleaccount": getattr(session, "isRole", False),
-        "metadata": getattr(session, "metadata", {}),
+        "dn": session.dn,
+        "fullname": session.fullname,
+        "email": session.email,
+        "isMember": session.is_member,
+        "isChair": session.is_chair,
+        "isRoot": session.is_root,
```
alitheg: Did we say we didn't want to keep this on the session now? Or is that a possible change for another time?

atr/server.py
```diff
@@ -474,35 +475,45 @@ async def bind_request_context_vars() -> None:
...
+                await asfquart.session.aclear()
+                sessions.invalidate_cache()
+                raise base.ASFQuartException("Account is disabled", errorcode=401)
+
+            admin_uid = session.admin_uid
+            if isinstance(admin_uid, str) and admin_uid and (not await ldap.is_active(admin_uid)):
+                log.auth_failure("oauth", "account_deleted_or_banned", admin_uid)
+                await asfquart.APP.sessions.revoke_by_uid(admin_uid)
+                await asfquart.session.aclear()
+                sessions.invalidate_cache()
```
alitheg: These 3 (or possibly 4 including the exception) lines could be extracted as I can see 4 times they're used identically

atr/post/user.py
sbp: Oh yes, I forgot about that. Only `is_root`, right?

atr/server.py
sbp: Good idea, thanks. Coming up in the next commit.

atr/post/user.py
alitheg: I think so - there are probably other things there we don't care about but that one is the most theoretically sensitive

atr/blueprints/common.py
sbp: We discussed this and we've agreed to comment it out, which I've done in the latest commit to this PR, f0a8978728d0ce874ca73b1b3742036fa09a728c.

Issue #1107, opened 2026-04-03
innerHTML Read Used Where textContent Is Appropriate


asf-tooling:
**ASVS Level(s):** L1

**Description:**

### Summary
The filter function in `projects-directory.js` reads `.innerHTML` instead of `.textContent` to extract the project name for filtering. Since `.card-title` contains an `<a>` tag, this causes the filter to match against HTML attributes (href, class) rather than just the visible project name. This is a functional bug rather than a direct XSS risk since it's a read operation, but demonstrates incorrect API usage.

### Details
The issue exists in `atr/static/js/src/projects-directory.js` line 26. The code reads innerHTML when it should read textContent for filtering purposes.

### Recommended Remediation
Replace innerHTML read with textContent:

```javascript
// In atr/static/js/src/projects-directory.js, line 26
// Change:
// const name = nameElement.innerHTML;

// To:
const name = nameElement.textContent;
```

This correctly reads only visible text as done in `committee-directory.js`.

### Acceptance Criteria
- [ ] innerHTML replaced with textContent
- [ ] Unit tests verify filter matches only visible text
- [ ] Unit tests verify filter does not match HTML attributes
- [ ] Manual testing confirms filter behavior is correct
- [ ] Code review confirms consistent API usage across JavaScript files

### References
- Source reports: L1:3.2.2.md
- Related findings: None
- ASVS sections: 3.2.2
- CWE: CWE-79

### Priority
Low

---

PR #1174, opened 2026-04-14
Project filter use textContent and not innerHTML plus cheap redos protection

### LINKED ISSUE #989: User Input Used Directly as RegExp Without Escaping in Project Directory Filter
**ASVS Level(s):** [L2-only]

**Description:**

### Summary
User input from the project filter textbox is passed directly to `new RegExp()` without escaping special characters, allowing regex metacharacters to be interpreted as pattern syntax rather than literal characters. This creates a ReDoS vulnerability where patterns like `(a+)+` can cause catastrophic backtracking and browser unresponsiveness. Invalid regex characters (e.g., `[`, `(`) cause unhandled exceptions, breaking the filter functionality entirely. Users expecting literal text search get unexpected wildcard behavior (e.g., `.` matches any character).

### Details
**Affected Files and Lines:**
- `atr/static/js/src/projects-directory.js:25-31` - RegExp without escaping

User input is used directly as a regex pattern without escaping special characters, allowing ReDoS and unexpected behavior.

### Recommended Remediation
Apply escaping to all regex special characters before constructing the RegExp object:

```javascript
const escapedFilter = projectFilter.replaceAll(/[.*+?^${}()|[\]\\]/g, '\\$&');
const regex = new RegExp(escapedFilter, 'i');
```

**Alternative:** Use `String.includes()` for simple text search instead of regex:

```javascript
const lowerFilter = projectFilter.toLowerCase();
projectRows.forEach(row => {
    const projectName = row.dataset.projectName.toLowerCase();
    row.style.display = projectName.includes(lowerFilter) ? '' : 'none';
});
```

### Acceptance Criteria
- [ ] Regex escaping implemented OR
- [ ] String.includes() used instead
- [ ] ReDoS prevented
- [ ] Invalid characters handled
- [ ] Literal text search works

### References
- Source reports: L2:1.2.9.md, L2:1.3.3.md
- Related findings: FINDING-212
- ASVS sections: 1.2.9, 1.3.3

### Priority
Medium

---

dave2wave:
closes: #989 and #1107

### DIFF
--- a/atr/static/js/src/projects-directory.js
+++ b/atr/static/js/src/projects-directory.js
@@ -20,9 +20,9 @@
 function filter() {
 	const projectFilter = document.getElementById("project-filter").value;
 	if (projectFilter.length > 100) {
-	    throw new Error("Input too long");
+		throw new Error("Input too long");
 	}
-	const regex = new RegExp(projectFilter, 'i');
+	const regex = new RegExp(projectFilter, "i");
 	const cards = document.querySelectorAll(".page-project-card");
 	let visibleCount = 0;
 	for (const card of cards) {
--- a/atr/static/js/src/projects-directory.js
+++ b/atr/static/js/src/projects-directory.js
@@ -19,13 +19,17 @@
 
 function filter() {
 	const projectFilter = document.getElementById("project-filter").value;
+	if (projectFilter.length > 100) {
+	    throw new Error("Input too long");
+	}
+	const regex = new RegExp(projectFilter, 'i');
 	const cards = document.querySelectorAll(".page-project-card");
 	let visibleCount = 0;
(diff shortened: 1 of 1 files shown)


sbp (commented):
The audit tool is wrong about this being a ReDoS vulnerability, because a user cannot attack themselves. But its point about users perhaps not expecting a pattern here may be valid. We should remove the unnecessary protection, and either add a UI note that this is a regular expression (potentially with error messages when the syntax is wrong), or switch, as the audit tool suggested, to a substring search. I have a slight preference for substring search, and have added commit 980eb428d6350fc1e64b47fa0519c0d153f57950 implementing that.

dave2wave:
Agreed. Merging.

PR #1176, opened 2026-04-16
#997 and #1022 - rework extraction and quarantine logic to rely on exarch

### LINKED ISSUE #997: Tar Archive Extraction Uses Explicitly Insecure Default Filter
**ASVS Level(s):** [L2-only]

**Description:**

### Summary
The `extract_member()` method uses `tar_filter='fully_trusted'` as its default parameter value. Python's PEP 706 and official documentation explicitly identify this filter as insecure. The `fully_trusted` filter allows absolute paths, path traversal sequences (`../`), device nodes, symlinks pointing outside extraction directory, and setuid/setgid bits. While mitigating controls exist (pre-extraction validation, quarantine workflow), the insecure default violates the principle of secure defaults.

### Details
Affected location: `atr/tarzip.py` lines 130-154

ASVS 1.5.2 states: "Deserialization mechanisms that are explicitly defined as insecure must not be used with untrusted input." The `fully_trusted` filter is explicitly documented as insecure by Python.

Mitigating controls:
- Pre-extraction validation in `check_archive_safety()`
- Quarantine workflow with `SecurityConfig`

However, the insecure default creates risk if these controls are bypassed or misconfigured.

### Recommended Remediation
Change the default `tar_filter` parameter from `'fully_trusted'` to `'data'` which is the secure default per PEP 706:

```python
def extract_member(
    self,
    member: tarfile.TarInfo,
    path: str,
    tar_filter: str = 'data',  # Changed from 'fully_trusted'
) -> int:
```

Update the docstring to:
- Document the security implications of each filter option
- Explicitly state that `fully_trusted` should only be used for verified trusted archives
- Add test cases to verify path traversal sequences are sanitized, absolute paths are converted to relative, and external symlinks are blocked or made safe

### Acceptance Criteria
- [ ] Default tar_filter is changed to 'data'
- [ ] Docstring documents security implications
- [ ] Test cases verify secure extraction behavior
- [ ] Path traversal and absolute paths are blocked by default

### References
- Source reports: L2:1.5.2.md
- Related findings: None
- ASVS sections: 1.5.2

### Priority
Medium

---

### LINKED ISSUE #1022: Pre-Extraction Safety Checks Do Not Verify Total Uncompressed Size
**ASVS Level(s):** [L2-only]

**Description:**

### Summary
The `check_archive_safety()` function runs BEFORE extraction and iterates every archive member with access to member.size attribute, but does not accumulate or validate total uncompressed size against MAX_EXTRACT_SIZE. Total size enforcement is deferred to the extraction phase via exarch SecurityConfig or streaming checks in archives.py. For ZIP files, `zipfile.ZipFile.infolist()` returns all member metadata from the central directory without decompressing any content, making pre-extraction size validation trivially achievable. An attacker could upload a ZIP with 50,000 files of 40 KB each (~2 GB total) that passes safety checks and begins extraction, consuming significant disk I/O and temporary storage before limits are enforced during extraction.

### Details
Affected locations:
- `atr/detection.py` lines 62-75: check_archive_safety() iterates members
- `atr/tasks/quarantine.py` lines 250-265: Calls check_archive_safety()

The function iterates all archive members and accesses member.size but never accumulates total size or validates against MAX_EXTRACT_SIZE.

### Recommended Remediation
Add total uncompressed size validation to `check_archive_safety()` by accumulating member.size during iteration and checking against `config.get().MAX_EXTRACT_SIZE`:

```python
def check_archive_safety(archive_path: Path, max_extract_size: int) -> list[str]:
    """Check archive for safety issues before extraction."""
    errors = []
    total_size = 0
    
    # ... existing iteration code ...
    for member in archive.infolist():
        total_size += member.size
        
        # Check against limit
        if total_size > max_extract_size:
            errors.append(
                f"Total uncompressed size {total_size} exceeds "
                f"MAX_EXTRACT_SIZE {max_extract_size}"
            )
            break  # Stop iteration, already over limit
        
        # ... existing per-member checks ...
    
    return errors
```

This prevents extraction from starting when size limits would be violated.

### Acceptance Criteria
- [ ] Total uncompressed size is validated before extraction starts
- [ ] Archives exceeding size limit are rejected before extraction
- [ ] Validation happens in pre-extraction safety check
- [ ] Test cases verify size limit enforcement

### References
- Source reports: L2:5.2.3.md
- Related findings: FINDING-242
- ASVS sections: 5.2.3

### Priority
Medium

---

alitheg:
Switch existing archive extraction to exarch and use the same `SecurityConfig` for those and quarantine extraction. Skip pre-extraction checking as it requires handling the archive without exarch, and doesn't give us anything additional to the extraction itself (provided we use the `SecurityConfig`, which we do). 

One thing to note:

I added some tests to confirm current behaviour before changing
- This raised the fact that the non-exarch extraction code skipped `._` files silently (intentionally), but quarantine doesn't. Since the only usage of that extraction code was now the sbom generator, I removed this behaviour

If all good, closes #997 and closes #1022.


atr/detection.py
```diff
@@ -16,15 +16,12 @@
...
 from typing import Final
 
 import puremagic
 
 import atr.models.attestable as models
 import atr.models.safe as safe
-import atr.tarzip as tarzip
 
 # TODO: Widen the range of types checked here
 QUARANTINE_ARCHIVE_SUFFIXES: Final[tuple[str, ...]] = (".tar.gz", ".tgz", ".zip")
```
dave2wave, line 27: These suffixes are not consistent the later one is missing .zip.

Let's make this TODO into an issue and cover all archives like bz2.

alitheg: Raised as  #1177

Issue #1177, opened 2026-04-16
Cover more types of archive in quarantine code


alitheg:
Detection currently only quarantines `".tar.gz", ".tgz", ".zip"`. Expand this to cover all archive types (eg. `bz2`)

dave2wave:
Sometimes archives are composed like `.tar.gz`

Here is a list of additional archive types:

```
.bz2
.xz
.whl
.war
.jar
.nar
```

You will need to see if `exarch` understands these.

sbp:
From [`detect.rs`](https://github.com/bug-ops/exarch/blob/main/crates/exarch-core/src/formats/detect.rs):

```rust
    let ext_lower = extension.to_ascii_lowercase();
    match ext_lower.as_str() {
        "tar" => Ok(ArchiveType::Tar),
        "gz" | "tgz" => {
            if let Some(stem) = path.file_stem()
                && stem.to_string_lossy().ends_with(".tar")
            {
                return Ok(ArchiveType::TarGz);
            }
            Ok(ArchiveType::TarGz)
        }
        "bz2" | "tbz" | "tbz2" => Ok(ArchiveType::TarBz2),
        "xz" | "txz" => Ok(ArchiveType::TarXz),
        "zst" | "tzst" => Ok(ArchiveType::TarZst),
        "zip" => Ok(ArchiveType::Zip),
        "7z" => Ok(ArchiveType::SevenZ),
        _ => Err(ExtractionError::UnsupportedFormat),
    }
```

So it won't understand some of these other types.

dave2wave:
See #553 for an inventory of file suffices.

asf-tooling:
**Automated triage** — analyzed at `main@2da7807a`

**Type:** `bug_fix`  •  **Classification:** `actionable`  •  **Confidence:** `medium`
**Application domain(s):** `artifact_upload`

### Summary
[...]

PR #1179, opened 2026-04-17
#1177 - allow zip format archives (jar, war, apk, vsix, whl, nar, nbm) to be quarantine checked

alitheg:
Adds a function to the quarantine process to create a temporary hard link with the `.zip` extension, for those file formats we know to be zip formatted internally. Exarch can then unzip them for safety checking.

I haven't added support to list the contents like we do for zip etc. - that would be essentially trivial using the same method if we wanted to.

I've added some tests but have also downloaded a file of each type and ensured they trigger the quarantine step.

### DIFF
--- a/atr/analysis.py
+++ b/atr/analysis.py
@@ -26,6 +26,7 @@ import sys
 from typing import Final
 
 ARTIFACT_SUFFIXES: Final[list[str]] = [
+    "apk",
     "bin",
     "crate",
     "deb",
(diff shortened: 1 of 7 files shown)


sbp:
Could we instead submit a PR to exarch either allowing those extensions, or allowing extensions to be configured?

alitheg:
Converted to draft in case there's some problem with them accepting a PR

alitheg:
https://github.com/bug-ops/exarch/pull/150

alitheg:
PR merged - awaiting next release

alitheg:
Not required with upstream contribution

PR #1181, opened 2026-04-17
Fix env var for expected secret crashes server

### LINKED ISSUE #948: Starting server with env var for expected secret crashes server
I (wrongly) tried to add PUBSUB config in environment variables and the server failed to start here:

```
        if key in repo_ini:
            value = repo_ini[key]
            return cast(value)
```

I checked, and even though my secrets file has no PUBSUB_URL in it, `"PUBSUB_URL" in repo_ini` returns True

dave2wave:
closes #948 

I noticed that configured secrets take precedence over environment variables. Perhaps it should be the other way around.

### DIFF
--- a/atr/config.py
+++ b/atr/config.py
@@ -49,12 +49,11 @@ def _config_secrets_get(
         if allow_not_found is False:
             raise
     else:
-        if key in repo_ini:
-            try:
-                value = repo_ini[key]
-                return cast(value)
-            except KeyError:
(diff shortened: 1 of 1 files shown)


sbp:
@dave2wave Amended to take out the existence check. The problem was that contains and getitem don't actually align, so it's not a great API.

alitheg (approved):

PR #1186, opened 2026-04-21
Project delete / archival

### LINKED ISSUE #510: Ensure that a project can only be deleted or archived under certain conditions, and that the state is clear
Part of #242 

- A project that has no releases may be deleted.
- A project that has no current and no archived releases may be deleted. Any candidate releases are also deleted.
- A project that has current or archived releases may only be archived. If archived then any current releases are also archived.
- If a project is the only project attached to a committee then it may neither be deleted nor archived.

alitheg:
I'll squash commits on this branch when it's done

### DIFF
--- a/atr/api/__init__.py
+++ b/atr/api/__init__.py
@@ -22,6 +22,7 @@ from typing import Any, Final, Literal
 import aiofiles.os
 import asfquart.base as base
 import openpgp
+import pydantic
 import quart
 import quart_rate_limiter as rate_limiter
 import quart_schema
@@ -1870,6 +1871,69 @@ def _issuer_matches_component(
     fingerprint_matches = (not issuer_fingerprints) or (fingerprint in issuer_fingerprints)
     key_id_matches = (not issuer_key_ids) or (key_id in issuer_key_ids)
     return fingerprint_matches and key_id_matches
+
+
+class TestProjectStatusArgs(pydantic.BaseModel):
+    project_key: safe.ProjectKey
+
+
+@api.typed
+@api.auth.public
+async def test_activate_project(
+    _test_activate_project: Literal["test/activate-project"],
+    data: TestProjectStatusArgs,
+) -> DictResponse:
+    """
+    URL: POST /test/activate-project
+
+    Restore a project to ACTIVE status (test mode only).
+    """
+    if not config.is_test_mode():
+        return quart.abort(404)
+    async with db.session() as session:
+        project = await session.project(key=str(data.project_key)).get()
+        if not project:
+            return quart.abort(404)
+        project = await session.merge(project)
+        project.status = sql.ProjectStatus.ACTIVE
+        await session.commit()
+    return {"ok": True, "project_key": str(data.project_key)}, 200
+
+
+@api.typed
+@api.auth.public
+async def test_archive_project(
+    _test_archive_project: Literal["test/archive-project"],
+    data: TestProjectStatusArgs,
+) -> DictResponse:
+    """
+    URL: POST /test/archive-project
+
+    Set a project to RETIRED status (test mode only).
+    """
+    if not config.is_test_mode():
+        return quart.abort(404)
+    async with db.session() as session:
+        project = await session.project(key=str(data.project_key)).get()
+        if not project:
+            return quart.abort(404)
+        project = await session.merge(project)
+        project.status = sql.ProjectStatus.RETIRED
+        await session.commit()
+    return {"ok": True, "project_key": str(data.project_key)}, 200
+
+
+def _committee_keys_path(committee: sql.Committee) -> safe.StatePath:
+    downloads_dir = paths.get_downloads_dir()
+    if committee.is_podling:
+        return downloads_dir / "incubator" / committee.key / "KEYS"
+    return downloads_dir / committee.key / "KEYS"
+
+
+def _committee_keys_url(host: str, committee: sql.Committee) -> str:
+    if committee.is_podling:
+        return f"https://{host}/downloads/incubator/{committee.key}/KEYS"
+    return f"https://{host}/downloads/{committee.key}/KEYS"
 
 
 def _jwt_asf_uid() -> str:
--- a/atr/get/projects.py
+++ b/atr/get/projects.py
@@ -17,6 +17,7 @@
 
 from __future__ import annotations
 
+from collections import Counter
 from typing import Literal
 
 import asfquart.base as base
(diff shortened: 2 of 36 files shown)


atr/api/__init__.py
```diff
@@ -1661,6 +1665,54 @@ async def vote_tabulate(
     ).model_dump(mode="json"), 200
 
 
+class _TestProjectStatusArgs(pydantic.BaseModel):
```
sbp: Whoops. We never prefix classes with an underscore, but I notice that our code conventions don't presently reflect that. (They only mention type variables.) I'll update the conventions.

tests/unit/test_archived_project_writer_guards.py
```diff
@@ -0,0 +1,326 @@
...
+
+
+def _async_context_manager(value: object) -> mock.MagicMock:
+    ctx = mock.MagicMock()
+    ctx.__aenter__ = mock.AsyncMock(return_value=value)
+    ctx.__aexit__ = mock.AsyncMock(return_value=None)
+    return ctx
+
+
+# --- revision --------------------------------------------------------------
```
sbp: Section comments break the function ordering script. Perhaps the function ordering script should be fixed to accommodate these, rather than removing all comments everywhere, though!

sbp: But note also that although these tests are nicely group-prefixed, they're also in entirely the wrong order.

tests/unit/test_create_revision.py
```diff
@@ -111,6 +110,7 @@ async def test_clone_from_older_revision_skips_merge_without_intervening_change(
     release = mock.MagicMock()
     release.phase = sql.ReleasePhase.RELEASE_PREVIEW
     release.project = mock.MagicMock()
+    release.project.status = sql.ProjectStatus.ACTIVE
```
sbp, line 114: These tests now fail:

```
tests/unit/test_create_revision.py ..F.F........                                                                        [ 27%]
```

With problems in `test_intervening_revision_with_path_provenance_verifies_sources` and `test_modify_path_provenance_flows_into_write_files_data`.

atr/storage/writers/project.py
```diff
@@ -186,17 +186,55 @@ async def create(self, committee_key: safe.CommitteeKey, display_name: str, labe
             else:
                 raise
 
+    async def archive(self, project_key: safe.ProjectKey) -> None:
```
sbp, line 214: This function doesn't seem to have any protections. What if there's a release in progress, e.g. holding a vote or something? Shouldn't the function hold a lock (or do some complex atomic query) and ensure that there are no active releases?

I notice it doesn't actually say that in https://github.com/apache/tooling-trusted-releases/issues/510 but it would seem like it would cause a problem if a project were archived when it was mid-release.

atr/storage/writers/project.py
```diff
@@ -186,17 +186,55 @@ async def create(self, committee_key: safe.CommitteeKey, display_name: str, labe
...
+        project = await self.__data.project(key=str(project_key), status=sql.ProjectStatus.ACTIVE).get()
+
+        if not project:
+            raise storage.AccessError(f"Project '{project_key}' not found.")
+
+        if project.committee_key:
+            committee_projects = await self.__data.project(
+                committee_key=project.committee_key, status=sql.ProjectStatus.ACTIVE
```
sbp, line 235: This correctly checks that the `status` is `ACTIVE`, but e.g. `category_add` and `category_remove` in this file don't have those checks. I may have forgotten to add the checks elsewhere, too.

atr/get/projects.py
```diff
@@ -94,28 +95,50 @@ async def add_project(
...
-            form_classes=".d-inline-block.m-0",
-            submit_classes="btn-sm btn-outline-danger",
-            submit_label="Delete project",
-            empty=True,
-            defaults={"project_key": str(project.key)},
-            confirm="Are you sure you want to delete this project? This cannot be undone.",
-        )
+        projects = await data.project(_committee=True, _releases=True).order_by(sql.Project.name).all()
+
+    committee_project_counts: Counter[str] = Counter(str(p.committee.key) for p in projects if p.committee)
```
sbp: Does this need to count `ACTIVE` projects only? Do we even need to track inactive projects in the database, necessarily? Perhaps we could have a different table for them or something.

atr/util.py
```diff
@@ -382,6 +360,16 @@ def email_from_uid(uid: str) -> str | None:
     return None
 
 
+def ensure_project_active(project: sql.Project) -> None:
```
sbp: This function is only ever used in `storage`, and imports `storage`, so I think it would be happier in `storage` somewhere. Probably just at the root is fine, or whatever makes sense.

tests/unit/test_archived_project_writer_guards.py
```diff
@@ -0,0 +1,326 @@
...
+
+# --- checks ----------------------------------------------------------------
+
+
+@pytest.mark.asyncio
+async def test_checks_ignore_add_propagates_chokepoint_block() -> None:
+    data = mock.MagicMock()
+    data.project = mock.MagicMock(return_value=_query_returning(_retired_project()))
+    writer = object.__new__(checks.CommitteeMember)
+    writer._CommitteeMember__data = data
```
sbp: Presumably only one or the other is needed? (`CommitteeMember` vs `CommitteeParticipant`)

playwright/test.py
```diff
@@ -573,6 +579,46 @@ def test_all(page: Page, credentials: Credentials, skip_slow: bool) -> None:
     logging.info(f"Tests took {round(finish - start, 3)} seconds")
 
 
+def test_archive_01_banner_absent_when_active(page: Page, credentials: Credentials) -> None:
```
sbp: Not sure we necessarily need Playwright tests for this feature. Could they be e2e tests? Also, if these crash half way through, then I think we get stuck with a `RETIRED` project and then all later runs will break. (This is why I had to add all the other state resetting stuff in these tests.)

tests/e2e/tokens/helpers.py
```diff
@@ -26,7 +26,7 @@
 def delete_token_by_label(page: Page, label: str) -> None:
     row = get_token_row_by_label(page, label)
     if row.count() > 0:
-        row.get_by_role("button", name="Delete").click()
```
sbp: Why did these get renamed? I don't see the corresponding change elsewhere.

sbp (changes requested):

atr/util.py
alitheg: Ah, yes - I looked at putting it in the root but the only things in the root currently were classes for the actual readers and writers and the methods you call to access them. If you're happy with essentially a utility class living in there too it's an easy move!

playwright/test.py
alitheg: Ah I thought we were trying to only use Playwright for new tests and e2e was old. Or have I got that entirely the wrong way around (or totally wrong)?

tests/e2e/tokens/helpers.py
alitheg: I thought Revoke was more correct for tokens, since unlike SSH keys (which are always *valid* it's just the user chooses where they're accepted), we issue them. I can leave at delete if you prefer.

alitheg: I made it in the SSH key changes set but because the e2e tests aren't run in CI (see my above comment/question about which tests are *current*), I missed that the tests were failing. Usually I run them all but clearly missed this time.

atr/get/projects.py
alitheg: A reasonable question. Was the idea that you can archive if there are live releases? In which case I assume we'd need to keep a hold of the project since the release depends on it?

playwright/test.py
sbp: Heh, yeah, it's the other way around! The e2e tests are better for new things because we can more easily run subsets of what we add there. We try to keep Playwright working and compatible only, although it should also absorb any particularly significant features.

tests/e2e/tokens/helpers.py
sbp: Ah, okay, no problem. I was only wondering why the tests changed but the code didn't!

atr/get/projects.py
sbp: I was just thinking about it being consistent with its handler, as part of the "everything should have `ACTIVE` checks" phase of the review. I think the idea is the opposite, that we _can't_ archive if there are ongoing releases.

alitheg: Dave's original comment read "A project that has current or archived releases may only be archived. If archived then any current releases are also archived." which I thought implied we should just allow archiving with current releases.

dave2wave: Sorry that I've change my direction.

alitheg: > Sorry that I've change my direction.

No problem! I'd just posted that before I brought it up in the call - thought it best to make sure we're on the same page. New direction makes most sense I think.

atr/get/vote.py
```diff
@@ -678,14 +678,15 @@ async def _render_vote_authenticated(
 
     potency = binding_word if is_binding else non_binding_word
     _render_binding_status(page, is_binding, binding_committee, vote_round)
-    _render_vote_delivery(page, archive_url, vote_recipient, trusted_vote=False)
+    _render_vote_delivery(page, release, archive_url, vote_recipient, trusted_vote=False)
```
sbp: The code may have changed in the meantime or something, because the actual function argument order is now `archive_url, release`.

atr/storage/writers/project.py
```diff
@@ -186,17 +188,59 @@ async def create(self, committee_key: safe.CommitteeKey, display_name: str, labe
...
+        if not project:
+            raise storage.AccessError(f"Project '{project_key}' not found.")
+
+        # TODO: once release archival is implemented, relax this further
+        post_draft_phases = {
+            sql.ReleasePhase.RELEASE_CANDIDATE,
+            sql.ReleasePhase.RELEASE_PREVIEW,
+            sql.ReleasePhase.RELEASE,
+        }
+        if any(r.phase in post_draft_phases for r in project.releases):
```
sbp, line 228: So this means that a project can be archived when it has draft releases? If so, it appears that the `delete` function in `atr.storage.writers.release` can't be used to later delete those drafts, because it now checks `storage.ensure_project_active(release.project)`. Therefore, if a project is archived when it has draft releases, those releases will be stranded.

sbp: In general, what was the intent behind allowing a release to be archived if it has draft releases? I remember we discussed this, but I don't remember what the outcome was.

atr/storage/writers/revision.py
```diff
@@ -394,6 +394,10 @@ async def clear_quarantine(
         version_key: safe.VersionKey,
         quarantined_id: int,
     ) -> None:
+        project = await self.__data.project(key=str(project_key)).demand(
+            storage.AccessError(f"Project '{project_key}' not found.")
+        )
+        storage.ensure_project_active(project)
```
sbp, line 400: The `set_tag` method in this module is missing `storage.ensure_project_active`. Ideally we'd make this the default, then opt out if we want to allow the action for non-active projects, but I'm not sure how to best go about that.

sbp (changes requested):

atr/get/vote.py
alitheg: Ah I may have messed up a rebase.

atr/storage/writers/project.py
alitheg: The original idea I think was that draft releases would be discarded if a project is archived (archived == retired).

PR #1207, opened 2026-04-28
Api auth decorators 1169

### LINKED ISSUE #1169: Add explicit authentication level decorators for API endpoint functions
Could be something like:

```
@api.auth.public
@api.auth.bearer
@api.auth.trusted_publisher
@api.auth.pat
```

Then, there should be tests checking when a route does not have any explicit decorator. That way we don't have the "fail open" problem where a route that we add without an authentication decorator is automatically public.

This issue is derived from the second half of #1040.

andrewmusselman:
## Add explicit authentication level decorators for API endpoints

Closes #1169.

Every `@api.typed` API route now declares its authentication level via exactly one of four decorators, and forgetting the decorator fails at import time rather than silently defaulting to public access.

```
@api.auth.public      # no authentication
@api.auth.bearer      # ATR-issued JWT in Authorization header
@api.auth.body_oidc   # Trusted Publisher OIDC token in request body
@api.auth.pat         # Personal Access Token (jwt_create only)
```

The `typed()` decorator inspects the wrapped function for an `_api_auth_level` marker and raises `TypeError` if it's missing. Adding a new route without an auth decorator means the server will not start.

### Three commits

Each commit is independently reviewable; the split mirrors how I built it.

**1. Scaffolding (no behavior change.)** Adds `atr/blueprints/api_auth.py` with the four decorators (all markers at this stage), wires `auth` onto the existing `api` namespace, adds the import-time enforcer in `typed()`, and marks all 46 existing API routes with their correct level. `@jwtoken.require` and the existing inline body-token validation stay where they are. The only new failure mode is "route registered without auth decorator → `TypeError`".

**2. Bearer consolidation.** `@api.auth.bearer` now wraps `jwtoken.require` and applies the `BearerAuth` OpenAPI security scheme itself, so every bearer endpoint can drop two redundant decorators. 36 lines deleted across 18 endpoints. No behavior change — same auth check, one decorator instead of three. Also drops the now-unused `import atr.jwtoken` from `atr/api/__init__.py`.

**3. body_oidc validation, registry, tests.**
- `@api.auth.body_oidc` pre-validates the Trusted Publisher OIDC token and exposes the verified state via `api.auth.trusted_publisher_context()` (a frozen `TrustedPublisherContext` dataclass on `quart.g`). All 7 body_oidc handlers migrated off direct JWT re-verification — zero calls to `interaction.trusted_jwt*` remain in the API module.
- New `interaction.trusted_project_for_payload()` and `trusted_release_for_payload()` helpers take an already-verified payload and do only the phase/project lookup. Existing `trusted_jwt()` and `trusted_jwt_for_dist()` are preserved and now share a `_trusted_dist_lookup()` helper.
- `tests/unit/api_auth_registry.yaml` records the canonical URL → auth level mapping for all 46 endpoints; a drift test asserts the live Quart route map matches.
- `tests/unit/test_api_auth_decorators.py` covers: marker correctness, import-time enforcement, stacking rejection, the registry drift, behavioral 401/200 per level, the `quart.g.tp_context` handoff, and the new `interaction` helpers.
- `docs/authentication-security.html` updated.

### Endpoint audit

Every API route is marked. Final distribution:

| Level | Count | Notes |
|---|---|---|
| `public` | 20 | No authentication |
| `bearer` | 18 | All previously `@jwtoken.require` |
| `body_oidc` | 7 | Trusted Publisher OIDC handlers |
| `pat` | 1 | `jwt_create` (PAT exchange) |

`@api.auth.pat` is intentionally a pure marker. The only consumer, `jwt_create`, *is* the PAT exchange endpoint, so wrapping its own logic in a decorator would just duplicate storage-layer code.

### Scope

API routes only, per discussion on the issue. Admin routes are admin-gated at the blueprint level, which is the right place for them. GET/POST web routes could plausibly benefit from a similar pattern in the future, but that's out of scope here.

### Edge cases worth knowing about

- `publisher_distribution_record` still tries `TrustedProjectPhase.FINISH` and falls back to `COMPOSE`. That fallback is handler logic, not auth, so it stays in the handler.
- `update_distribution_task_status` requires `ctx.asf_uid is None` (ATR-driven workflow only). Also handler logic. If this pattern spreads, a `body_oidc_atr_only` variant would be reasonable.
- `distribute_ssh_register` and `distribution_record_from_workflow` use `trusted_release_for_payload()` because they take `asf_uid`/`project_key`/`version_key` from the body and need the dist-style lookup.
- Rate limiting stays orthogonal to auth.
- `/api/openapi.json` doesn't go through `@api.typed` and is filtered explicitly in the coverage test.

### Verification

- `make unit`, `make e2e`, `make check`
- Rebased on `main`

### DIFF
--- a/atr/blueprints/api_auth.py
+++ b/atr/blueprints/api_auth.py
@@ -47,8 +47,12 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Final, Literal
 
+import quart_schema
+
+import atr.jwtoken as jwtoken
+
 if TYPE_CHECKING:
-    from collections.abc import Awaitable, Callable
+    from collections.abc import Awaitable, Callable, Coroutine
     from typing import Any
 
 
@@ -60,15 +64,27 @@ AUTH_LEVEL_ATTR: Final[str] = "_api_auth_level"
 VALID_LEVELS: Final[frozenset[AuthLevel]] = frozenset({"public", "bearer", "body_oidc", "pat"})
 
 
-def bearer[F: Callable[..., Awaitable[Any]]](func: F) -> F:
-    """Marker for routes authenticated by an ATR-issued Bearer JWT.
+def bearer[**P, R](
+    func: Callable[P, Coroutine[Any, Any, R]],
+) -> Callable[P, Awaitable[R]]:
+    """Require an ATR-issued Bearer JWT in the ``Authorization`` header.
 
-    PR 1: marker only. Handlers still declare ``@jwtoken.require`` and
-    ``@quart_schema.security_scheme([{"BearerAuth": []}])`` explicitly.
-    PR 2 folds both of those into this decorator and removes the duplicate
-    stack across the ~18 bearer endpoints.
+    Folds in two concerns that previously had to be declared separately
+    on every bearer endpoint:
+
+    1. ``@jwtoken.require`` — validates the JWT and populates
+       ``quart.g.jwt_claims`` so handlers can read the caller's identity
+       via ``_jwt_asf_uid()``.
+    2. ``@quart_schema.security_scheme([{"BearerAuth": []}])`` — advertises
+       the scheme in the generated OpenAPI document.
+
+    The auth-level marker is applied to the outermost wrapper so
+    :func:`atr.blueprints.api.typed` can read it back.
     """
-    return _mark("bearer", func)
+    wrapped = jwtoken.require(func)
+    wrapped = quart_schema.security_scheme([{"BearerAuth": []}])(wrapped)
+    _mark("bearer", wrapped)
+    return wrapped
 
 
 def body_oidc[F: Callable[..., Awaitable[Any]]](func: F) -> F:
--- a/atr/blueprints/api_auth.py
+++ b/atr/blueprints/api_auth.py
@@ -1,0 +1,120 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+"""Explicit authentication level decorators for API endpoints.
+
+Every route decorated with :func:`atr.blueprints.api.typed` must also be
+decorated with exactly one of the auth level decorators in this module.
+The :func:`atr.blueprints.api.typed` decorator enforces this at import
+time; forgetting the decorator raises :class:`TypeError` and the server
+will not start.
+
+Usage::
+
+    @api.typed
+    @api.auth.bearer
+    @quart_schema.validate_response(models.api.FooResults, 200)
+    async def foo(...):
+        ...
+
+Decorator order: ``@api.typed`` on the outside, then ``@api.auth.<level>``,
+then any ``@quart_schema`` or ``@rate_limiter`` decorators closer to the
+function.
+
+The levels are:
+
+- ``public``     No authentication. Route may be called by anyone.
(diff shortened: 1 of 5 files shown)


atr/blueprints/api_auth.py
```diff
@@ -0,0 +1,232 @@
...
+
+        payload, asf_uid = await interaction.validate_trusted_jwt(publisher, jwt)
+        quart.g.tp_context = TrustedPublisherContext(
+            payload=payload,
+            asf_uid=asf_uid,
+            publisher=publisher,
+        )
+        return await func(*args, **kwargs)
+
+    wrapped = quart_schema.security_scheme([{"TrustedPublisherJwt": []}])(wrapper)
```
sbp: Does the `TrustedPublisherJwt` security scheme exist? There's no other reference to it in the code.

atr/blueprints/api_auth.py
```diff
@@ -0,0 +1,232 @@
...
+        if (not isinstance(publisher, str)) or (not isinstance(jwt, str)):
+            raise exceptions.BadRequest(
+                "Trusted Publisher auth requires 'publisher' and 'jwt' string fields in the request body"
+            )
+
+        # Import lazily to avoid a top-level cycle: atr.db.interaction
+        # depends on a lot of things, some of which depend on the blueprints.
+        import atr.db.interaction as interaction
+
+        payload, asf_uid = await interaction.validate_trusted_jwt(publisher, jwt)
```
sbp: The docstring says:

>     On failure (bad signature, wrong audience, unsupported publisher,
>     unknown actor) a 401 ``ASFQuartException`` is raised before the
>     handler body executes, so handlers never see unauthenticated calls.

But this call can fail in all sorts of ways and those will produce 500 errors, rather than 401 errors.

sbp (commented):
Also:

* We normally don't add docstrings, or one line at most. (This might be a mistake, but we should review it and decide together.)
* Type ignores shouldn't be necessary in tests, as tests are ignored.
* Top level comments break the function order script. Try running `git diff --cached --name-only -- '*.py' | xargs -I {} sh scripts/fix_order.sh {}` on staged changes.

andrewmusselman:
@sbp Note: pyright fails on atr/storage/writers/keys.py:136 (dsa_bits attribute access on PublicParamsInfo). This is pre-existing on main after 9cc1dbbe Restore support for DSA keys — rpgp-py 0.19.7's PublicParamsInfo doesn't expose dsa_bits at runtime. Unrelated to this PR.

andrewmusselman:
> * Type ignores shouldn't be necessary in tests, as tests are ignored.
> 
> * Top level comments break the function order script. Try running `git diff --cached --name-only -- '*.py' | xargs -I {} sh scripts/fix_order.sh {}` on staged changes.



Good notes — all three addressed. Docstrings collapsed to one line per symbol. The two `# type: ignore[no-untyped-def]` comments in `tests/unit/test_api_auth_decorators.py` are gone. The four explanatory top-level comments I'd added (in `api.py` above `__all__ and _route_auth_levels`, in `api_auth.py` above `AUTH_LEVEL_ATTR`) are removed; `_route_auth_levels` moved up next to the other `_BLUEPRINT*` module state so `fix_order.py` is happy. Verified clean with `python3 scripts/fix_order.py` on every touched file.

atr/blueprints/api_auth.py
andrewmusselman: Right — that scheme was never registered with `QuartSchema(...)` in `server.py`, only `BearerAuth` is. I dropped the dangling `quart_schema.security_scheme([{"TrustedPublisherJwt": []}])` call from `body_oidc`. Body OIDC routes now have no security annotation in the OpenAPI doc. Registering a proper scheme — `OpenIdSecurityScheme` against GitHub's OIDC discovery URL is probably the closest fit, since OpenAPI doesn't have a native body-located scheme — is worth a follow-up issue but felt out of scope here.

atr/blueprints/api_auth.py
andrewmusselman: You're right, the 401 claim was aspirational. I made the docstring true: `body_oidc` now catches `interaction.InteractionError` (unsupported publisher), `pyjwt.InvalidTokenError` (bad signature, expired, malformed JWT), and `pydantic.ValidationError` (payload doesn't match `TrustedPublisherPayload` schema) from `validate_trusted_jwt(...)`, and converts each to `ASFQuartException(errorcode=401)`. Pre-existing `ASFQuartException` is re-raised unchanged so 502s for genuine upstream failures (TLS errors fetching JWKS, etc.) keep their status. Body validation errors (missing data, missing publisher/jwt fields) also map to 401. Two affected unit tests updated.

andrewmusselman:
@sbp Dropped api_auth_registry.yaml and test_api_auth_registry_matches_live_routes.

PR #1221, opened 2026-05-05
Project cycle model - data models and initial CLE doc generation

alitheg:
It would be good to get this side live, so that going forward new projects/releases have the right data in place. What do you think?

### DIFF
--- a/atr/cle.py
+++ b/atr/cle.py
@@ -1,0 +1,293 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+"""ECMA-428 Common Lifecycle Enumeration document generator.
+
+Reads from `LifecycleEvent` rows and renders to ECMA-428 wire format.
+Internal event names map to spec event types via `_SPEC_TYPE`:
+
+    release  -> released
+    archive  -> endOfDistribution
+    eod      -> endOfDevelopment
+    eos      -> endOfSupport
+    eol      -> endOfLife
+    withdraw -> withdrawn
+
+Per ECMA-428 section 7.9, a `withdrawn` event retracts the *publication*
+of a previously-emitted event - typically a correction. The withdrawn
+event itself and the targeted event both remain in the document; the
+withdrawal references its target via the `eventId` field. ATR mostly uses it
+for corrections to previously-published cycle dates (see
+policy.edit_cycle_dates).
+
+We don't emit endOfMarketing, supersededBy, or componentRenamed.
+"""
+
+from __future__ import annotations
+
+import datetime
+from typing import TYPE_CHECKING, Any, Final
+
+import atr.models.sql as sql
+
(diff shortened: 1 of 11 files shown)


atr/cle.py
```diff
@@ -0,0 +1,237 @@
...
+# Versioned schema URL. ECMA-428 is still draft, so this points at the
+# tc54 working copy and may need updating when the spec ships.
+CLE_SCHEMA_URL: Final[str] = "https://ecma-tc54.github.io/ECMA-428/cle.v1.0.0.schema.json"
+
+# CLE event types we currently emit. The spec defines five more we don't:
+# endOfSupport, withdrawn, endOfMarketing, supersededBy, componentRenamed.
+_RELEASE_EVENT: Final[str] = "released"
+_END_OF_DISTRIBUTION_EVENT: Final[str] = "endOfDistribution"
+_END_OF_DEVELOPMENT_EVENT: Final[str] = "endOfDevelopment"
+_END_OF_MAINTENANCE_EVENT: Final[str] = "endOfMaintenance"
```
sbp: I don't see `endOfMaintenance` in [ECMA-428 § 7](https://ecma-tc54.github.io/ECMA-428/#sec-event-types), or anywhere else in the specification.

atr/cle.py
```diff
@@ -0,0 +1,237 @@
...
+
+    releases_by_cycle = _releases_by_cycle(releases)
+    for cycle in cycles:
+        cycle_releases = releases_by_cycle.get(cycle.cycle_key, [])
+        for column, event_type in _CYCLE_EVENT_MAP:
+            date_value = getattr(cycle, column)
+            if date_value is not None:
+                raw.append(_cycle_event(project, cycle_releases, event_type, date_value))
+
+    # Sort ascending by effective date so id assignment is stable for a given
```
sbp: From [ECMA-428 § 9](https://ecma-tc54.github.io/ECMA-428/#sec-event-processing-rules):

> **Immutability**: Once an event is published, it cannot be modified. New events must be added to correct or update information. [...] **ID Assignment**: Event IDs must be assigned as auto-incrementing integers in the order events are added (not by effective date).

But this block of code wouldn't be stable in the current code if we added one in the middle, like an event that we forgot about, or was published by a background task, etc. Might be too hard to fix in the current way that we're doing it, so we'd need a rule to ensure that we can never insert backdated events.

atr/cle.py
```diff
@@ -0,0 +1,237 @@
...
+
+    This is a derived view, not a spec form. The document still has the
+    component-level shape ECMA-428 prescribes, but only includes events
+    touching this release: its own released/archived events, plus the
+    lifecycle events of the cycle it belongs to.
+    """
+    return _document(project, [cycle], [release], now=now)
+
+
+def _cycle_event(
```
sbp: https://ecma-tc54.github.io/ECMA-428/#sec-event-type-endofdevelopment
https://ecma-tc54.github.io/ECMA-428/#sec-event-type-endofsupport

These also require a `supportId` attribute.

atr/cle.py
```diff
@@ -0,0 +1,237 @@
...
+    applies to all versions in the (currently empty) cycle. With releases, we
+    emit a literal pipe-delimited list of every version, which is correct
+    regardless of version_method but verbose for large cycles. Once semver
+    projects exist, this can return a proper range like
+    `vers:semver/>=2.0.0|<3.0.0` derived from cycle_match.
+    """
+    scheme = _vers_scheme(project)
+    if not cycle_releases:
+        return f"vers:{scheme}/*"
+    constraints = "|".join(f"={r.version}" for r in cycle_releases)
```
sbp: Looking at [ECMA-428 § Annex A](https://ecma-tc54.github.io/ECMA-428/#sec-vers), it appears that the literal version syntax is just the version with no prefix. There's no `=1.2.3` in there.

sbp (changes requested):

atr/cle.py
alitheg: > Event IDs must be assigned as auto-incrementing integers in the order events are added (not by effective date).

Aha. This could be the argument for having a CLE events table...

atr/cle.py
alitheg: I'd imagine that would be a static thing per project

Issue #913, opened 2026-03-19
Project reference metadata


dave2wave:
## Project reference metadata

field | description
--------|--------
lifecycle page | A link to the PMC web page which describes support plans / lifecycle for releases.
download page | The link to the required download page. 
repository | Project git/svn repository checkout urls.
homepage | Project website.
sort description | Short project description
bug database | URL to one of these systems: Bugzilla, JIRA, or GH Issues
mailing lists | Page on the project website
standards | List of implemented standards urls

asf-tooling:
**Automated triage** — analyzed at `main@2da7807a`

**Type:** `new_feature`  •  **Classification:** `actionable`  •  **Confidence:** `medium`
**Application domain(s):** `project_committee_management`

### Summary
Issue #913 requests adding project reference metadata fields (homepage, repository, download page, bug database, mailing lists, lifecycle page, standards) to ATR's project model. @gstein noted this looks like DOAP data and asked about storage; @dave2wave confirmed and pointed to issue #139 for context. The external data source (`ProjectStatus` in `atr/datasources/apache.py`) already parses most of these fields from projects.apache.org, but `_update_projects()` currently discards them. The implementation requires new DB columns on `sql.Project`, persisting fetched data, and adding UI/forms for viewing/editing these fields.

### Where this lives in the code today

#### `atr/datasources/apache.py` — `ProjectStatus` (lines 192-209)
_currently does this_
Already parses homepage, repository, bug_database, download_page, mailing_list, shortdesc, and implements (standards) from the external DOAP data source — but these are not persisted to the database.

```python
class ProjectStatus(schema.Strict):
    category: list[str] = schema.factory(list)
    created: str | None = None
    description: str | None = None
    programming_language: list[str] = schema.Field(alias="programming-language", default_factory=list)
    doap: str | None = None
    homepage: str
    name: str
    pmc: str | None
    shortdesc: str | None = None
    repository: list[str | dict] = schema.factory(list)
    release: list[Release] = schema.factory(list)
    bug_database: str | None = schema.alias_opt("bug-database")
    download_page: str | None = schema.alias_opt("download-page")
    license: str | None = None
    mailing_list: str | None = schema.alias_opt("mailing-list")
    maintainer: list[MaintainerInfo] = schema.factory(list)
    implements: list[ImplementsInfo] = schema.factory(list)
```

#### `atr/storage/writers/project.py` — `CommitteeMember` (lines 77-85)
_extension point_
[...]

PR #1229, opened 2026-05-07
Draft PR for cycle UI

alitheg:
Adds UI for managing cycle dates to the project settings, and moves other settings to a new tabbed interface

### DIFF
--- a/atr/blueprints/common.py
+++ b/atr/blueprints/common.py
@@ -249,7 +249,9 @@ async def flash_form_error(form_cls: type, error: pydantic.ValidationError) -> A
     summary = form.flash_error_summary(errors, flash_data)
     await quart.flash(summary, category="error")
     await sessions.form_error_put(quart.request.path, flash_data)
-    return quart.redirect(quart.request.path)
+    # Persist the query string if required
+    redirect_url = quart.request.full_path if quart.request.query_string else quart.request.path
+    return quart.redirect(redirect_url)
 
 
 async def parse_body(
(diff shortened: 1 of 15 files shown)


alitheg:
Screenshot_20260508_172850
Screenshot_20260508_172835
Screenshot_20260508_172827

dave2wave:
For better looking tabs at the `ul` level switch from `nav-tabs` to `nav-pills`

alitheg:
@dave2wave Updated the screenshots with the new tabs, fields from #913 added and releases moved automatically when I changed my versioning scheme

alitheg:
@sbp @dave2wave I've also reimplemented the CLE generation logic into a new place that could theoretically be extracted as a python library for managing (parsing, building) CLE documents - I didn't see any existing library for this and it could be a good thing to contribute? See other branch 795cdaa6

alitheg:
Oh one more @sbp - on the date field question. I made it ISO format as requested, but then we lose the date picker because it's not a browser date field any more. It looks like the display field if we _do_ use a date field is locale dependent, so you and I would see DD-MM-YYYY and Dave and Andrew would see MM-DD-YYYY. Do you think better to keep ISO and have manual entry or have the date picker?

dave2wave:
> Oh one more @sbp - on the date field question. I made it ISO format as requested, but then we lose the date picker because it's not a browser date field any more. It looks like the display field if we _do_ use a date field is locale dependent, so you and I would see DD-MM-YYYY and Dave and Andrew would see MM-DD-YYYY. Do you think better to keep ISO and have manual entry or have the date picker?

I would revert the ISO format change. Locale specific behavior for picking dates is desirable and accessible.

PR #1232, opened 2026-05-07
Drop dsa_bits reference; rpgp-py 0.19.7 doesn't expose it

andrewmusselman:
`make check` fails pyright on main with reportAttributeAccessIssue for PublicParamsInfo.dsa_bits, added in 9cc1dbbe (Restore support for DSA keys). rpgp-py 0.19.7 has no dsa_bits attribute and no equivalent field; for a DSA key, all numeric fields on PublicParamsInfo return None, only kind == dsa identifies it.

DSA keys now raise a clear ValueError instead of AttributeError. RSA and ECC unaffected. Restoring DSA bit-length validation needs upstream work in rpgp-py.

## Testing

`make check` passes

## Rebase info

```
$ git fetch upstream main
From github.com:apache/tooling-trusted-releases
 * branch              main       -> FETCH_HEAD
$ git rebase upstream/main
Current branch fix-keys-dsa-bits is up to date.
```

### DIFF
--- a/atr/storage/writers/keys.py
+++ b/atr/storage/writers/keys.py
@@ -133,16 +133,19 @@ def _key_length(key: openpgp.PublicKey) -> int:
     rsa_bits = public_params.rsa_bits
     if isinstance(rsa_bits, int):
         return rsa_bits
-    dsa_bits = public_params.dsa_bits
-    if isinstance(dsa_bits, int):
-        return dsa_bits
     curve_bits = public_params.curve_bits
     if isinstance(curve_bits, int):
         return curve_bits
-    raise ValueError(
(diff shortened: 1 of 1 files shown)


sbp:
> rpgp-py 0.19.7 has no dsa_bits attribute and no equivalent field

No, [it does](https://github.com/xelandernt/rpgp-py/blob/87bf51c023487f77251a6b78b4937b96ced269df/src/openpgp/_openpgp.pyi#L264-L265):

```python
    @property
    def dsa_bits(self) -> Optional[int]: ...
```

I know this because [I added it](https://github.com/xelandernt/rpgp-py/pull/7).

Issue #139, opened 2025-06-09
Allow Projects to be defined via .asf.yaml


dave2wave:
One way that we could improve the UX around defining projects is to add a section to the `.asf.yaml` process which would allow one or more projects that are sourced in one of a PMC's repositories.

This could look something like:

```
projects:
  - name: project1
     id: ...
     folder: ...
     description: ...
     keys: ...
     tags: ...
     doap: (url or generate)
  - name: project2
     ...
```

The .asf.yaml processing could then use an ATR api to create/update the project and generate (or use) the DOAP. This ATR api could also update `projects.apache.org`'s `projects.xml`.

dave2wave:
To be discussed with #138

gstein:
All the svn projects get mirrored to GitHub, and the `.asf.yaml` facility is available to them. I know of several svn-based projects that use `.asf.yaml` and even GHA workflows.

dave2wave:
On #138 this reference was provided: [fields of DOAP](https://projects.apache.org/create.html). I think that this should give most of the fields for the yaml:

[fields of DOAP](https://projects.apache.org/create.html) defines these fields:

Field | Form value | Example pattern
-- | -- | --
Project name | Apache | Apache [Committee] [Subproject]
Project homepage | https:// | https://[committee].apache.org/[subproject]
Short Description |  
Long Description | 
Categories | Big Data  ...
Programming Languages | ActionScript 
Bug Database URL |   | https://github.com/apache/[committee]-[subproject]/issueshttps://issues.apache.org/jira/projects/[JIRA-ID]/
Download Page URL |   | https://[committee].apache.org/[subproject]/download.html
Implemented Standard | Title, Body. Consortium, ID, URL

This can be refactored some for the yaml.

dave2wave:
To be worked on as a part of #147. Converting to a work issue.

potiuk:
Nice!

dave2wave:
It is currently deferred, but it will be done. The .asf.yaml features are Infra managed and added here: https://github.com/apache/infrastructure-asfyaml/tree/main/asfyaml/feature

dave2wave:
When implemented this should allow for every project metadata field and policy fields to be set from `.asf.yaml` - to help with that there should be an endpoint that returns a current yaml for the project's current configuration.

sebbASF:
How does ATR know where to find the .asf.yaml files? Is there an index somewhere?

dave2wave:
[...]

PR #1238, opened 2026-05-12
#921 - add endpoint to create/update project data

### LINKED ISSUE #462: Projects don't have a chair
https://github.com/apache/tooling-trusted-releases/blob/b5a7cda8fe31d44e228d9051b6791ddd99b80bc1/atr/datasources/apache.py#L210

### LINKED ISSUE #465: PMCs have a charter; projects don't
If this is where PMCs are described, then it is missing the charter, which is taken from the board resolution (currently stored in committee-info.yaml):

https://github.com/apache/tooling-trusted-releases/blob/b5a7cda8fe31d44e228d9051b6791ddd99b80bc1/atr/datasources/apache.py#L80

Also AFAIK, projects don't have a charter:
https://github.com/apache/tooling-trusted-releases/blob/b5a7cda8fe31d44e228d9051b6791ddd99b80bc1/atr/datasources/apache.py#L211

### LINKED ISSUE #921: Define yaml/json format api for project metadata, cycle, and policy settings
With an agreed format for overall project metadata we can add the api for post/get of a project data and implement #139

alitheg:
Schema built from policy fields and (now) complete metadata. The endpoint accepts partial models and updates in-place if the project exists already.

Added datasource changes too. PR closes #462, #465, #921

### DIFF
--- a/atr/api/__init__.py
+++ b/atr/api/__init__.py
@@ -858,6 +858,39 @@ async def policy_update(
 
 
 @api.typed
+@jwtoken.require
+@quart_schema.security_scheme([{"BearerAuth": []}])
+@quart_schema.validate_response(models.api.ProjectConfigResults, 200)
+async def project_config_upsert(
+    _project_config: Literal["project/config"],
+    data: models.api.ProjectConfigArgs,
+) -> DictResponse:
+    """
+    URL: POST /project/config
+
+    Upsert a project's full configuration.
+
+    Creates a project if it on by the specified key does not yet exist.
+    Otherwise, updates all specified fields for an existing project.
+    The caller must be a committee member of the specified committee,
+    and for an existing project committee_key must match the current value.
+    """
+    asf_uid = _jwt_asf_uid()
+    try:
+        async with storage.write_as_committee_member(str(data.committee_key), asf_uid) as wacm:
+            created = await wacm.project.upsert_config(data)
+    except storage.AccessError as e:
+        raise _http_exception_from_storage_access_error(e) from e
+    except ValueError as e:
+        raise exceptions.BadRequest(str(e))
+    return models.api.ProjectConfigResults(
+        endpoint="/project/config",
+        success=True,
+        created=created,
+    ).model_dump(mode="json"), 200
+
+
+@api.typed
 @quart_schema.validate_response(models.api.ProjectGetResults, 200)
 async def project_get(
     _project_get: Literal["project/get"],
--- a/atr/datasources/apache.py
+++ b/atr/datasources/apache.py
@@ -41,6 +41,7 @@ import atr.util as util
 _WHIMSY_COMMITTEE_INFO_URL: Final[str] = "https://whimsy.apache.org/public/committee-info.json"
 _WHIMSY_COMMITTEE_RETIRED_URL: Final[str] = "https://whimsy.apache.org/public/committee-retired.json"
 _WHIMSY_PROJECTS_URL: Final[str] = "https://whimsy.apache.org/public/public_ldap_projects.json"
+_PROJECTS_COMMITTEE_URL: Final[str] = "https://projects.apache.org/json/foundation/committees.json"
 _PROJECTS_PROJECTS_URL: Final[str] = "https://projects.apache.org/json/foundation/projects.json"
 _PROJECTS_PODLINGS_URL: Final[str] = "https://projects.apache.org/json/foundation/podlings.json"
 _PROJECTS_GROUPS_URL: Final[str] = "https://projects.apache.org/json/foundation/groups.json"
@@ -81,6 +82,20 @@ class User(schema.Strict):
 
 
 class Committee(schema.Strict):
+    chair: str
+    charter: str
+    established: str
+    group: str
+    homepage: str
+    id: str
+    name: str
+    rdf: str
+    reporting: int | None = None
+    roster: Annotated[list[User], helpers.DictToList(key="id")]
+    shortdesc: str
+
+
+class WhimsyCommittee(schema.Strict):
     name: str
     display_name: str
     site: str | None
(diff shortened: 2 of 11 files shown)


atr/api/__init__.py
```diff
@@ -857,6 +857,69 @@ async def policy_update(
...
+) -> DictResponse:
+    """
+    URL: POST /project/config
+
+    Upsert a project's full configuration.
+
+    Creates a project if it on by the specified key does not yet exist.
+    Otherwise, updates all specified fields for an existing project.
+    The caller must be a committee member of the specified committee,
+    and for an existing project committee_key must match the current value.
```
dave2wave, line 876: When we implement #139 we will need to share a PAT with GitBox to allow asfyaml to access this api.

atr/api/__init__.py
```diff
@@ -857,6 +857,69 @@ async def policy_update(
...
+            f"Project '{data.project_key}' belongs to committee '{existing.committee_key}', not '{data.committee_key}'"
+        )
+    try:
+        async with storage.write_as_committee_member(str(data.committee_key), asf_uid) as wacm:
+            if existing is None:
+                if (data.project is None) or (data.project.name is None) or (not data.project.name.strip()):
+                    raise exceptions.BadRequest(
+                        f"Project '{data.project_key}' does not exist; project.name is required to create it"
+                    )
+                await wacm.project.create(
```
sbp: This is calling `wacm` three times, but ideally it would be atomic. Perhaps there could be another `wacm.project` method that does all three of these at once, with a `begin_immediate`?

atr/models/api.py
```diff
@@ -377,11 +381,78 @@ def validate_policy_fields(self) -> "PolicyUpdateArgs":
...
+    version_pattern: str | None = None
+    cycle_match: str | None = None
+    branch_template: str | None = None
+
+    @pydantic.field_validator("version_method", mode="before")
+    @classmethod
+    def version_method_to_enum(cls, v):
+        if isinstance(v, str):
+            try:
+                return sql.VersionMethod.__members__[v]
```
sbp: But for example we also have:

```python
    @pydantic.field_validator("vote_mode", mode="before")
    @classmethod
    def vote_mode_to_enum(cls, v):
        if isinstance(v, str):
            try:
                return sql.VoteMode(v)
            except ValueError:
                raise ValueError(f"'{v}' is not a valid VoteMode")
        return v
```

And that's our pattern throughout, so I think this should return `sql.VersionMethod(v)`, and use a `ValueError` exception branch.

atr/api/__init__.py
```diff
@@ -857,6 +857,69 @@ async def policy_update(
...
+
+    Creates a project if it on by the specified key does not yet exist.
+    Otherwise, updates all specified fields for an existing project.
+    The caller must be a committee member of the specified committee,
+    and for an existing project committee_key must match the current value.
+    """
+    asf_uid = _jwt_asf_uid()
+    sections_applied: list[Literal["project", "policy"]] = []
+    created = False
+    async with db.session() as session:
```
sbp: Maybe this existence and filtering stuff belong in the storage interface, in the new wrapper function that I propose in the R894 comment.

atr/models/safe.py
```diff
@@ -24,11 +24,11 @@
 
 import pydantic
 
-_ALPHANUM: Final = frozenset(string.ascii_letters + string.digits + "-")
+_ALPHANUM: Final = frozenset(string.ascii_letters + string.digits + "-+")
```
sbp: This is used by committee key too, but I don't think we should allow `+` in those.

atr/storage/writers/project.py
```diff
@@ -235,6 +237,53 @@ async def edit_metadata(self, form: shared.projects.EditMetadataForm) -> None:
...
 
+    async def edit_project_partial(
+        self,
+        project_key: safe.ProjectKey,
+        args: api.ProjectConfigProjectArgs,
+    ) -> None:
+        provided = args.model_fields_set
+        if not provided:
+            return
+        str_fields = {
```
sbp: These are all `form.OptionalURL` in the web form (`atr.shared.projects.EditMetadataForm`), which does validation. In the API, these are just `str`, so I guess they're not validated the same way?

atr/datasources/apache.py
```diff
@@ -384,7 +411,10 @@ def _project_status(pmc: sql.Committee, project_key: str, project_status: Projec
 
 
 async def _update_committees(
-    data: db.Session, ldap_projects: LDAPProjectsData, committees_by_name: Mapping[str, Committee]
+    data: db.Session,
+    ldap_projects: LDAPProjectsData,
+    whimsy_committees_by_name: Mapping[str, WhimsyCommittee],
+    commiteess: dict[str, Committee],
```
sbp: Typo: `commiteess`

atr/datasources/apache.py
```diff
@@ -409,9 +439,12 @@ async def _update_committees(
         committee.committers = project.members
         # We create PMCs for now
         committee.is_podling = False
-        committee_info = committees_by_name.get(name)
+        whimsy_info = whimsy_committees_by_name.get(name)
+        if whimsy_info:
+            committee.name = whimsy_info.display_name
+        committee_info = commiteess.get(name)
```
sbp: Typo: `commiteess`

atr/datasources/apache.py
```diff
@@ -81,6 +82,20 @@ class User(schema.Strict):
 
 
 class Committee(schema.Strict):
+    chair: str
```
sbp, line 85: The commit message talks about removing a project chair, but I'm not sure what that means. The only reference to a chair in the PR is this one.

sbp (changes requested):

atr/models/api.py
alitheg: Ah I copied this from DistributionPlatform. I'll make that match too.

atr/datasources/apache.py
alitheg: Ah I put it back of course - will amend.

atr/models/api.py
alitheg: Actually I'll raise that separately as it will need a change to the automation builds and that's not as high priority

PR #1243, opened 2026-05-14
#911 - add artifact metadata to data model, add artifact signature count to signing keys

### LINKED ISSUE #911: Artifact catalog model
The **_Release Catalog_** should include a table of release artifacts.

Track metadata related to current release artifacts

## Artifact catalog model

field   | description
------|--------------
artifact path | The release artifact.
key fingerprint | The gpg public key. (Not the path, not the signature, the public key.
signature path | The path, not a url.
checksum path | If multiple pick the "best"
classification  | result of classify process.
project key |
version key |
svn revision | revision where artifact was added to `dist:release` - the `commit` with either a `mv` or an `add`

Historically not all of a release's artifacts were added in the same revision.

### LINKED ISSUE #920: Reorganize committee page
The committee page should be reorganized into sections like the project page in #917 

1. Projects. Should become a list of project tiles like on committees page. Also, add a create a new project button if the user is a PMC member.
2. Signing keys.
    - Sort the table by asf_uid then fingerprint.
    - Add a column to show how many current artifacts are signed with that key (see #911 )
    - Add a column to indicate if the key is expired.
    - Add a column to indicate if the user is currently a member or committer of the committee.
3. Roster. Instead of a count of members and committers show a roster sorted by ASF UID of
    - ASF UID 
    - Name
    - PMC Member check box
    - Artifact signature check box (or count)

alitheg:
Thoughts on the data model?

I assume we'll create one entry in this per artifact we push to SVN, and remove them when releases are removed from SVN?

### DIFF
--- a/atr/get/committees.py
+++ b/atr/get/committees.py
@@ -19,6 +19,7 @@ import datetime
 from typing import Literal
 
 import asfquart.base as base
+import sqlalchemy
 import sqlmodel
 
 import atr.blueprints.get as get
@@ -67,6 +68,20 @@ async def view(session: web.Public, _committees: Literal["committees"], name: sa
         )
         names: dict[str, str | None] = {u.asfuid: u.name for u in user_rows.scalars().all()}
 
+        fingerprints = [k.fingerprint for k in committee.public_signing_keys]
+        artifact_counts: dict[str, int] = {}
+        if fingerprints:
+            via = sql.validate_instrumented_attribute
+            count_rows = await data.execute(
+                sqlalchemy.select(
+                    via(sql.Artifact.key_fingerprint),
+                    sqlalchemy.func.count(),
+                )
+                .where(via(sql.Artifact.key_fingerprint).in_(fingerprints))
+                .group_by(via(sql.Artifact.key_fingerprint))
+            )
+            artifact_counts = {fp: n for fp, n in count_rows.all() if fp is not None}
+
     project_list = list(committee.projects)
     signing_keys = sorted(
         committee.public_signing_keys,
@@ -85,6 +100,7 @@ async def view(session: web.Public, _committees: Literal["committees"], name: sa
         roster=roster,
         names=names,
         signing_keys=signing_keys,
+        artifact_counts=artifact_counts,
         algorithms=shared.algorithms,
         now=datetime.datetime.now(datetime.UTC),
         email_from_key=util.email_from_uid,
--- a/atr/models/sql.py
+++ b/atr/models/sql.py
@@ -1190,6 +1190,10 @@ class Release(sqlmodel.SQLModel, table=True):
         back_populates="release", sa_relationship_kwargs={"cascade": "all, delete-orphan"}
     )
 
+    # 1-M: Release -> [Artifact]
+    # M-1: Artifact -> Release
+    artifacts: list["Artifact"] = sqlmodel.Relationship(back_populates="release")
+
     # The combination of key and version must be unique
     __table_args__ = (sqlmodel.UniqueConstraint("project_key", "version", name="unique_project_version"),)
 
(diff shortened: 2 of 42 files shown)


dave2wave:
We will mark them archived when archived the catalog is permanent.

atr/get/committees.py
```diff
@@ -67,6 +68,20 @@ async def view(session: web.Public, _committees: Literal["committees"], name: sa
         )
         names: dict[str, str | None] = {u.asfuid: u.name for u in user_rows.scalars().all()}
 
+        fingerprints = [k.fingerprint for k in committee.public_signing_keys]
```
sbp, line 71: I guess this means we're doing a bit like we do for checks, where we say anything that hashes the same way is the same thing. I keep thinking that this might have been a mistake in checks, because then we can't easily e.g. delete results for a single release. But I'm still not sure either way. If it was a mistake there (if!), then it might be a mistake here too.

atr/models/sql.py
```diff
@@ -1641,6 +1641,31 @@ class ReleaseFileState(sqlmodel.SQLModel, table=True):
...
 
 
+# Artifact: Project, Release, Signing Key
+class Artifact(sqlmodel.SQLModel, table=True):
+    project_key: str = sqlmodel.Field(primary_key=True, foreign_key="project.key", **example("example"))
+    version: str = sqlmodel.Field(primary_key=True, **example("0.0.1"))
+    artifact_path: str = sqlmodel.Field(primary_key=True, **example("apache-example-0.0.1.tar.gz"))
+    # Link to ATR release record when one exists - null for historical SVN artifacts
+    release_key: str | None = sqlmodel.Field(
+        default=None, foreign_key="release.key", index=True, **example("example-0.0.1")
```
sbp: For these foreign keys, shouldn't they cascade deletions? There was a similar problem with the events recently that I fixed, which manifested during tests. Also, since the `Artifact` here _does_ seem to be scoped to a release, perhaps that underlines my other comment about fingerprints. Perhaps they should be clearly scoped in the UI queries to a release too?

migrations/versions/0083_2026.05.14_7c4f8a2e.py
```diff
@@ -0,0 +1,73 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+"""Add Artifact catalog model for issue 911
+
+Revision ID: 0083_2026.05.14_7c4f8a2e
+Revises: 0082_2026.05.14_a1b2c3d4
+Create Date: 2026-05-14 00:00:00.000000+00:00
```
sbp: Should we always use the real time, or are we relaxing that to just the date?

atr/get/committees.py
alitheg: I think on the checks front it would be a mistake if you didn't also have a reference for that release to all the checks, but we have that in attestable data so we should be able to find them - I suppose the issue in there is not being able to remove them as we don't know if they're used by any other release

alitheg: I thought in this case, on the committee page, the point was to show the number of signed artifacts in total?

alitheg:
> We will mark them archived when archived the catalog is permanent.

Ok - I've added the release as a join on artifacts so we can know if archived and therefore display them as such in whatever UI they're shown in.

atr/get/committees.py
dave2wave: Checks are unique to a release. Fingerprints span releases. We do want to track all uses of a fingerprint so that we can disallow deletion of that fingerprint.

alitheg:
Closing this as it'll be merged shortly but keeps triggering new builds

PR #1244, opened 2026-05-14
Cache a map of hashed email addresses to ASF UIDs in a file

sbp:
We needed a reverse map in the file too, for the updates.

### DIFF
--- a/atr/admin/__init__.py
+++ b/atr/admin/__init__.py
@@ -1294,7 +1294,7 @@ async def validate_jwt_post(
 
 
 async def _check_keys(fix: bool = False) -> str:
-    await cache.email_uid_view_or_live()
+    email_uid_lookup = await cache.email_uid_view_or_live()
     bad_keys = []
     async with db.session() as data:
         keys = await data.public_signing_key().all()
@@ -1304,7 +1304,7 @@ async def _check_keys(fix: bool = False) -> str:
                 uids.append(key.primary_declared_uid)
             if key.secondary_declared_uids:
                 uids.extend(key.secondary_declared_uids)
-            asf_uid = await util.asf_uid_from_uids(uids, use_ldap=False)
+            asf_uid = await util.asf_uid_from_uids(uids, email_uid_lookup, use_ldap=False)
             if asf_uid != key.apache_uid:
                 bad_keys.append(f"{key.fingerprint} detected: {asf_uid}, key: {key.apache_uid}")
             if fix:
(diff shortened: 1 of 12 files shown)


atr/cache.py
```diff
@@ -95,6 +134,153 @@ async def admins_startup_load() -> None:
...
+        await asyncio.to_thread(os.chmod, temp_path, 0o400)
+        await aiofiles.os.rename(temp_path, cache_path)
+    except Exception:
+        with contextlib.suppress(FileNotFoundError):
+            await aiofiles.os.remove(temp_path)
+        raise
+
+
+async def email_uid_startup_load() -> None:
+    cache_data = await _email_uid_read_from_file_async()
```
alitheg: Since we erase them on shutdown, surely there's no point trying to load them again? In fact, if we erase on startup is there any point in saving to a file at all?

atr/util.py
alitheg: I wonder if `asf_uid_from_email` should be renamed or made internal so that the cache version is the main place you can call to map an email?

atr/admin/__init__.py
```diff
@@ -1293,7 +1294,7 @@ async def validate_jwt_post(
 
 
 async def _check_keys(fix: bool = False) -> str:
-    email_to_uid = await util.email_to_uid_map()
+    await cache.email_uid_view_or_live()
```
alitheg: I see two patterns - "ask for the cache to make sure it's populated then call a method later that I know accesses the cache" (in this file) or "ask for the cache and return it so I can use it" (in keys.py and tabulate.py). Both work obviously - do you think we should have both or should we standardise on one?

atr/cache.py
sbp: It's not immediately obvious from the commit, but the reason why we need a file is because `atr.tasks.keys.import_file` (in a file _not_ changed in this commit) calls `wacm.keys.import_keys_file`, which calls `ensure_associated` which calls `__ensure` which calls `cache.email_uid_view_or_live()`. Because it's being called from a task, `asfquart.APP` is `None`, so it needs to read it from the file instead.

Loading on startup is a mistake, a remnant of earlier drafts which did not erase on shutdown. When I changed that, I didn't take out the load. Thanks for spotting it!

atr/util.py
sbp: Agreed, thanks. I'll make it internal since it's only used in `util.py` and is really only intended as a helper in there, not for use elsewhere in ATR.

atr/admin/__init__.py
sbp: Yeah, the latter is better. Going to standardise on that.

sbp:
All comments addressed in abb626f24438660d7077bdd288bf3b3a2c80efb8.

alitheg (approved):

PR #1248, opened 2026-05-18
473 auto archive release

alitheg:
Adds auto archive option. It must be set on a project's metadata to be enabled for a release.

On creating a release you can choose to auto-archive the previous release in the cycle when the new release is announced.


atr/post/announce.py
```diff
@@ -105,32 +96,16 @@ async def selected(
...
-    current_hash = construct.template_hash(subject_template)
-    if current_hash != announce_form.subject_template_hash:
-        return await session.form_error(
-            "subject_template_hash",
-            "The subject template has been modified since you loaded the form. Please reload and try again.",
-        )
+    # Re-resolve the prior release server-side rather than trusting any value
+    # the form might submit. The form's auto_archive field is the opt-in only.
+    archive_prior: sql.Release | None = None
+    if announce_form.auto_archive:
```
sbp: Shouldn't this also check the policy level option? The comment indicates that this actually happens, that the form isn't trusted by itself, but I can't find where the policy option is checked. It doesn't seem to be checked in `prior_release_for_archive`, which is quite a short function. There's also the `Release.archive_prior_release` property... what happens in all possible combinations? I assume that if the release flag is `False`, then no matter what we shouldn't archive. What if the release flag is `True` but the policy flag is `False`? Inconsistent state, error? So it should only archive if they're both `True`.

atr/post/announce.py
```diff
@@ -146,6 +121,14 @@ async def selected(
                 email_cc=announce_form.email_cc,
                 email_bcc=announce_form.email_bcc,
             )
+            if archive_prior is not None:
+                archive_error = await wacm.release.archive(project_key, archive_prior.safe_version_key)
```
sbp, line 125: This should probably be atomic with the release, like in a previous PR where I suggested wrapping them both in another storage interface method to make them atomic, but in this case it's a bit more complicated because we'll be writing to SVN and sending emails. At some point we may just have to accept that breakages may leave us in a partial state, and figure out how we can recover.

atr/storage/writers/release.py
```diff
@@ -1048,6 +1053,85 @@ def __init__(
...
+        release = await self.__data.release(
+            project_key=str(project_key),
+            version=str(version_key),
+            _committee=True,
+        ).get()
+        if release is None:
+            return f"Release {project_key!s} {version_key!s} not found"
+        if release.phase != sql.ReleasePhase.RELEASE:
+            return f"Release {project_key!s} {version_key!s} is not in the release phase"
+        if release.archived is not None:
```
sbp: TOCTOU here. I think this should go in the `sqlmodel.update` below.

atr/get/announce.py
```diff
@@ -248,25 +249,36 @@ async def _render_page(
         # Custom widget for download_path_suffix with custom documentation
         download_path_widget = _render_download_path_field(default_download_path_value, download_path_description)
 
+        prior_release_version = ""
+        if release.archive_prior_release:
+            prior = await interaction.prior_release_for_archive(release.project, release.version)
+            if prior is not None:
+                prior_release_version = prior.version
```
sbp, line 256: But the `auto_archive` checkbox will still be checked if `prior` is `None`? Should probably be contingent on `archive_prior_release` _and_ `prior` not being `None`.

atr/post/announce.py
alitheg: With the other plumbing to be added (SVN etc) I wonder if in this case it's worth leaving it here - especially since if that operation fails the user should be able to go and archive the release themselves.

Issue #938, opened 2026-03-24
Supporting the Erlang distribution channel


dave2wave:
From jonatan@maennchen.ch

Hi Dave,

Thanks, this is really useful context.

On the Erlang/Elixir side, we already have a few things that should fit in pretty well with what you’re building:

## SBOM generation

We maintain rebar3_sbom and mix_sbom, which generate CycloneDX SBOMs directly from builds:
https://github.com/erlef/rebar3_sbom
https://github.com/erlef/mix_sbom

I saw CouchDB has both a mix.exs and a rebar.config.script, so one of those should work without much trouble. Happy to sanity check the output against what you expect in the catalog.

## Source SBOM / analysis

We’re also using ORT: https://oss-review-toolkit.org/ort/
That gives us a source-level view, which might or might not be useful depending on how far you want to go beyond release artifacts.

## Trusted publishing

Not there yet for Hex.pm, but it’s something we’re working towards:
https://github.com/hexpm/hexpm/issues/1193

## Vulnerability handling (CNA)

The EEF runs a CNA covering Erlang, Elixir, and Hex:
https://cna.erlef.org/

We’re currently looking into a pre-disclosure / heads-up list for downstream users of Erlang. Since CouchDB ships Erlang, that might also be relevant on your side:
https://github.com/erlef-cna/website/pull/61

## EEF working groups

We also have a few active working groups in the EEF. One that might be particularly relevant here is the OpenRiak group. They are dealing with many of the same kinds of issues you will see in CouchDB as another Erlang-based database.

It would be great to share some of that experience, and more generally I would be very happy if we could bring the ASF and Erlang communities closer.

I think there’s a pretty natural overlap here, especially around SBOM generation and how artifacts are exposed. Let me know where it would make sense to align or collaborate.

asf-tooling:
**Automated triage** — analyzed at `main@2da7807a`

**Type:** `discussion`  •  **Classification:** `no_action`  •  **Confidence:** `high`
**Application domain(s):** `distribution_tracking`, `sbom_analysis`, `automated_checks`

### Summary
[...]

Issue #751, opened 2026-02-23
Investigate remote promotion of artifacts on third party platforms


sbp:
When staging versions are uploaded to third party distribution platforms such as DockerHub, MavenCental, or PyPI, we should consider allowing remote promotion of those artifacts to production releases, if possible. This is a minor performance improvement. It can't be counted as a reduction in complexity: if anything, it's an increase in complexity, because for platforms that don't support remote upgrade we will still have to support uploading from ATR again. The files are always on ATR, so they will never need to be re-fetched, which means that there are no performance gains there.

There are some caveats to this. We would have to track the exact revision of files that were uploaded in the staging version. We should do this anyway, of course, but then when allowing promotion we also have to check that the files haven't been updated in subsequent revisions. We have to check that the files being promoted are the same as those that were voted on. This is an extra check that we don't currently have to do in ATR because the artifacts in the finish phase are, by definition, those that have been voted on.

We also have to consider how the promotion works on each platform. This would replace uploading from ATR, so there wouldn't be any extra complexity there. From a quick browse of documentation, ArtifactHub doesn't support promotion because it [only records metadata](https://artifacthub.io/docs/topics/faq/#can-i-install-applications-directly-from-artifact-hub); DockerHub [only supports promotion if using mutable tags](https://docs.docker.com/docker-hub/repos/manage/hub-images/immutable-tags/); Maven Central [does support promotion](https://central.sonatype.org/publish/publish-portal-api/#publish-or-drop-the-deployment); npm supports [mutable tags that it calls dist-tags](https://docs.npmjs.com/adding-dist-tags-to-packages); and PyPI does not support promotion from Test PyPI to PyPI and we have received a report that projects use PyPI for staging anyway.

dave2wave:
I just received an email about Erlang. I think that we need to design an issue template and then we can collect the proper information about each distribution channel and cover everything we need. We probably need to revisit our schema.

andrewmusselman:
@dave2wave @sbp @alitheg draft here for review, I'm not sure how much of this should go in `atr/docs` since it's an investigation for a decision

# Distribution Platform Investigation

## Overview

This document investigates remote promotion capabilities across the distribution platforms ATR supports, and identifies what schema or model changes would be needed. The goal is to determine whether ATR can promote staged artifacts to production on each platform without re-uploading, and whether this is worth the added complexity.

## Platform Capability Matrix

[...]

PR #1264, opened 2026-05-22
Track the staging revision of a recorded distribution

andrewmusselman:
# Track the staging revision of a recorded distribution

Closes part of #751. Related: #938.

## Summary

Two small, independent changes to the distribution layer. Both fall out of andrewmusselman's RFC on #751 as "independently useful regardless of whether remote promotion is ever built."

**A nullable `staging_revision_key` foreign key on the `distribution` table.** This records which `Revision` was the source of a staged distribution. The original issue calls out the need directly: *"we have to check that the files being promoted are the same as those that were voted on."* Without a record of the connection between a staged distribution and the underlying revision, ATR has no way to enforce that — for local re-upload today, or for remote promotion later. The column is the foundation that any future verification check can stand on. It is nullable so existing rows and call sites that don't yet know the revision key remain valid; a `TODO(#751)` in `CommitteeMember.__upgrade_staging_to_final` marks where the assertion will live once call sites populate the field.

**Removal of the hard-coded `{ARTIFACT_HUB, PYPI, MAVEN}` whitelist in `shared.distribution._template_url`.** That set duplicated knowledge already encoded in `DistributionPlatformValue.template_staging_url`. The two had to be edited in lockstep, which is a quiet footgun: adding a new staging URL (for example, enabling the commented-out GITHUB platform in `sql.py`, or eventually a Hex platform for #938) would silently fail until someone remembered to update the parallel set. After this change the enum is the single source of truth — a platform supports staging iff its value declares a `template_staging_url`. The error message also now names the offending platform, which makes the failure mode obvious.

## Files changed

| File | Change |
| --- | --- |
| `atr/models/sql.py` | Add nullable `staging_revision_key: str \| None` FK column to `Distribution` |
| `atr/shared/distribution.py` | Remove `{ARTIFACT_HUB, PYPI, MAVEN}` whitelist in `_template_url`; the error message now names the platform |
| `atr/storage/writers/distributions.py` | Thread `staging_revision_key` through `record()` and `record_from_data()` as an optional pass-through; add `TODO(#751)` marker in `__upgrade_staging_to_final` |
| `migrations/versions/0086_2026.05.21_38b1cb95.py` | Alembic migration: nullable column + FK; reversible via `downgrade()` |
| `tests/unit/test_distribution_staging.py` | Six tests covering `_template_url` for every platform (regression-pinning that no whitelist gets reintroduced) and `Distribution.staging_revision_key` defaults |

## Smoke tests run

All four of these were exercised locally against `docker compose up --build` on this branch. Anyone reviewing can reproduce them in a few minutes.

**1. Schema and migration.** On server boot, ATR's automatic `alembic upgrade head` plus the follow-up `alembic check` (in `atr/db/__init__.py`) validates that the migration applies cleanly and that the resulting DB schema matches the SQLModel definitions. Boot succeeded with `Alembic check passed: DB schema matches models`. To re-verify locally: `docker compose up --build` and watch for that line in the logs.

**2. Whitelist removal does not regress staging for platforms that worked before.** Navigate to `/distribution/stage/record/<project>/<version>` and submit the form with a non-existent package on each of the three platforms that have a staging URL. The expected result is an upstream 404 (or similar parse error) — not the old whitelist message.

- **ArtifactHub** (`helm`, `asfowner`, `example`, `1`): `404 Not Found` from `https://staging.artifacthub.io/api/v1/packages/helm/asfowner/example/1` ✅
- **Maven Central** (`orgapacheexample`, `example`, `1`): `404 Not Found` from `https://repository.apache.org:4443/repository/maven-staging/orgapacheexample/example/maven-metadata.xml` ✅

(Maven groupIds are dotted in the real world, but the form's `owner_namespace` field is typed `safe.Alphanumeric` which rejects `.`. That's a pre-existing form-validation issue, unrelated to this PR, and worth filing separately.)

**3. New error message names the platform.** Submit the same form with **Docker Hub** (`library`, `nginx`, `latest`). Expected: `Platform Docker Hub does not provide a staging API endpoint.` ✅

The `RuntimeError` raised by `_template_url` propagates to a 500 page, but that propagation is pre-existing — the old whitelist branch raised the same `RuntimeError` and produced the same 500. Only the message text changed. The 500-wrapping itself is worth a separate UX issue.

**4. New column is present and behaves as expected.**

```shell
sqlite3 state/database/atr.db "PRAGMA table_info(distribution);" | grep staging_revision_key
```

returns the column, nullable, FK to `revision.key`. After recording any distribution through the UI, `SELECT staging_revision_key FROM distribution` returns `NULL` for every row — expected, since this PR deliberately does not wire call sites to populate the field. That's a follow-up.

## What this deliberately does not change

The RFC laid out several follow-ups and called for a "narrow rather than generic" implementation. This PR matches that posture.

- **No `supports_promotion` / `promotion_mechanism` columns.** Not needed until a second platform actually needs them. `supports_promotion: bool` is derivable from `promotion_mechanism is not None`, and `promoted: bool` / `promoted_at` duplicate state that the existing `staging` boolean flip already encodes.
- **No decision on the commented `GITHUB` enum.** That needs a human yes/no. The whitelist removal here means a future enable just works.
- **No automatic revision-key lookup at call sites.** The new parameter on `record()` and `record_from_data()` is optional and defaults to `None`. Wiring `post/distribution.py`, `tasks/distribution.py`, and `api/__init__.py` to look up the current `Revision` and pass it through is a small follow-up that becomes meaningful now that the column exists.
- **No verification assertion in `__upgrade_staging_to_final`.** The `TODO(#751)` marker shows where it lands; the assertion follows once call sites populate the field.
- **No Hex/Erlang (#938) work.** That's blocked upstream on `hexpm/hexpm#1193` and needs its own design. The whitelist removal does make a future Hex platform easier to add.
- **No investigation document in `atr/docs/`.** Better handled as a separate PR once the team decides where investigation docs live.

## Required acknowledgements

- [x] I have read CONTRIBUTING.md.
- [x] "Allow maintainer edits" is enabled.
- [x] No new dependencies are introduced.
- [x] No `# noqa` or `# type: ignore` added.
- [x] Commit message uses imperative mood and is under 72 chars on the first line.
- [x] `make check`, `sh tests/run-unit.sh` pass locally.
- [x] Migration round-trips cleanly (`alembic upgrade head` → `downgrade -1` → `upgrade head`).

### DIFF
--- a/atr/models/sql.py
+++ b/atr/models/sql.py
@@ -1556,6 +1556,12 @@ class Distribution(sqlmodel.SQLModel, table=True):
     api_url: str | None = sqlmodel.Field(default=None)
     web_url: str | None = sqlmodel.Field(default=None)
     created_by: str | None = sqlmodel.Field(default=None)
+    # Records which Revision was the source of the staged artifacts.
+    # Set when staging=True so that the finish-phase code can verify the
+    # artifacts being promoted match those that were voted on, addressing
+    # the caveat in https://github.com/apache/tooling-trusted-releases/issues/751
+    # The column is nullable so that pre-existing rows are unaffected.
+    staging_revision_key: str | None = sqlmodel.Field(default=None, foreign_key="revision.key")
     # The API response can be huge, e.g. from npm
     # So we do not store it in the database
     # api_response: Any = sqlmodel.Field(sa_column=sqlalchemy.Column(sqlalchemy.JSON))
--- a/atr/storage/writers/distributions.py
+++ b/atr/storage/writers/distributions.py
@@ -156,6 +156,7 @@ class CommitteeMember(CommitteeParticipant):
         upload_date: datetime.datetime | None,
         api_url: str | None = None,
         web_url: str | None = None,
+        staging_revision_key: str | None = None,
     ) -> tuple[models.sql.Distribution, bool]:
         release = await self.__data.release(key=str(release_key), _project=True).demand(
             storage.AccessError(f"Release '{release_key}' not found.")
@@ -178,6 +179,9 @@ class CommitteeMember(CommitteeParticipant):
             api_url=api_url,
             web_url=web_url,
             created_by=self.__asf_uid,
+            # Only meaningful for staging rows; ignore the value otherwise
+            # so we don't accidentally pin a production row to a revision.
+            staging_revision_key=staging_revision_key if staging else None,
         )
         if existing is None:
             self.__data.add(dist)
@@ -227,6 +231,7 @@ class CommitteeMember(CommitteeParticipant):
         staging: bool,
         dd: models.distribution.Data,
         allow_retries: bool = False,
+        staging_revision_key: str | None = None,
     ) -> tuple[models.sql.Distribution, bool, models.distribution.Metadata | None]:
         release = await self.__data.release(key=str(release_key), _project=True).demand(
             storage.AccessError(f"Release '{release_key}' not found.")
@@ -254,6 +259,7 @@ class CommitteeMember(CommitteeParticipant):
                         upload_date=None,
                         api_url=None,
                         web_url=None,
+                        staging_revision_key=staging_revision_key,
                     )
                     return dist, added, None
                 raise storage.AccessError(f"Failed to get API response from distribution platform: {error}", status=502)
@@ -278,6 +284,7 @@ class CommitteeMember(CommitteeParticipant):
             upload_date=upload_date,
             api_url=api_url,
             web_url=web_url,
+            staging_revision_key=staging_revision_key,
         )
         return dist, added, metadata
 
@@ -302,6 +309,12 @@ class CommitteeMember(CommitteeParticipant):
             version=version,
         ).demand(RuntimeError(f"Distribution {tag} not found"))
         if existing.staging:
+            # TODO(#751): When remote promotion of artifacts to third party
+            # platforms is implemented, assert here that existing.staging_revision_key
+            # equals the current Revision.key for the release before flipping
+            # staging to False, so that the artifacts being promoted are
+            # guaranteed to match those that were voted on. The column was
+            # added in #751 for exactly this purpose.
             existing.staging = False
             existing.pending = pending
             existing.upload_date = upload_date
--- a/tests/unit/test_distribution_staging.py
+++ b/tests/unit/test_distribution_staging.py
@@ -1,0 +1,114 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import pytest
+
+import atr.models.distribution as distribution
+import atr.models.safe as safe
+import atr.models.sql as sql
+import atr.shared.distribution as shared_distribution
+
+
+def _data(platform: sql.DistributionPlatform, *, owner_namespace: str | None = None) -> distribution.Data:
+    return distribution.Data(
+        platform=platform,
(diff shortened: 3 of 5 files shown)


atr/storage/writers/distributions.py
```diff
@@ -302,6 +309,12 @@ async def __upgrade_staging_to_final(
             version=version,
         ).demand(RuntimeError(f"Distribution {tag} not found"))
         if existing.staging:
+            # TODO(#751): When remote promotion of artifacts to third party
+            # platforms is implemented, assert here that existing.staging_revision_key
+            # equals the current Revision.key for the release before flipping
+            # staging to False, so that the artifacts being promoted are
+            # guaranteed to match those that were voted on. The column was
+            # added in #751 for exactly this purpose.
```
alitheg, line 317: I could be wrong (@sbp please confirm) - but when I was working in a related area yesterday I realised that moving to preview (vote) creates a new revision. You can't create a distribution from the vote page, so the staging revision will always be one behind the actual revision voted on (even though it's the same code). In some places in the code (for vote stage access to checks for example) we use the immediate previous revision. We could do that in the future check for this, potentially? But it's something to consider more broadly - the vote stage has a revision with little to no metadata attached. Equally - some projects won't vote on a distribution so we should make sure that any future check here is only mandatory if there *was* a staging distribution!

alitheg: Otherwise, I don't see any obvious problems here, but I'll wait for @sbp too!

tests/unit/test_distribution_staging.py
```diff
@@ -0,0 +1,114 @@
...
+    return distribution.Data(
+        platform=platform,
+        owner_namespace=safe.Alphanumeric(owner_namespace) if owner_namespace else None,
+        package=safe.Alphanumeric("example"),
+        version=safe.VersionKey("1.0.0"),
+        details=False,
+    )
+
+
+# --- _template_url: enum is now the single source of truth ---
```
sbp, line 36: These comments mean that we can't run `fix_order` on the file.

atr/storage/writers/distributions.py
sbp: A new revision is created at the end of the vote phase, when going to finish. This was historical, when we had revisions in the finish phase. The staging revision shouldn't be a revision behind during the vote, only afterwards. It wasn't a problem before when we had finish phase revisions that the new revisions weren't the same number as the one voted on. ATR ensured that the changes applied were meaning preserving.

sbp: Oh, and we can still remove RC tags and empty directories, so we still need this.

atr/models/sql.py
```diff
@@ -1556,6 +1556,12 @@ class Distribution(sqlmodel.SQLModel, table=True):
     api_url: str | None = sqlmodel.Field(default=None)
     web_url: str | None = sqlmodel.Field(default=None)
     created_by: str | None = sqlmodel.Field(default=None)
+    # Records which Revision was the source of the staged artifacts.
+    # Set when staging=True so that the finish-phase code can verify the
+    # artifacts being promoted match those that were voted on, addressing
+    # the caveat in https://github.com/apache/tooling-trusted-releases/issues/751
+    # The column is nullable so that pre-existing rows are unaffected.
+    staging_revision_key: str | None = sqlmodel.Field(default=None, foreign_key="revision.key")
```
sbp, line 1564: FK without a cascade.

sbp: Could set null on delete. Only happens if distributions aren't deleted first anyway, but then it's order dependent which would be nice to avoid.

sbp (commented):
Seems fine otherwise.

sbp:
I resolved the merge errors, but I am questioning the need for this, at least at present. This PR is a solution to the following requirement that I wrote in the original post of #751:

> We have to check that the files being promoted are the same as those that were voted on.

The solution in this PR is to check that there has been no intervening revision, but as @alitheg pointed out, that cannot ever work because we _always_ create a new revision after voting. There can also be subsequent revisions later in the finish phase.

We have three sets of files: 1. the staged third party distribution artifacts, 2. the artifacts in the vote revision on ATR, 3. the production release third party distribution artifacts. We'd like to know that all of these are the same. If they're not the same they _might_ still be secure, but we can't guarantee it through ATR. To know whether the files are the same, we need the hashes of the files. Some platforms make hashes available in metadata files, so a first step would be to figure out which ones make those available. Then we would need to figure out a workflow for this, because we need to know where we would perform this check. We also need to figure out what we want the results of the check to be.

Overall, though, I don't think that we should be working on this feature before we even have remote promotion implemented. That's the core, MVP of #751: allowing a user on ATR to press a button to promote an artifact remotely, if possible. The big questions are what platforms support this, which we've already investigated, and what credentials are required.

PR #1271, opened 2026-05-22
Warn about and then delete inactive unfinished releases

sbp:
@alitheg This is a draft, really, but I'd appreciate a review anyway!

This PR sends out warnings if a release has been inactive for 80 days. When it reaches 90 days, it deletes it. It makes an exception on the first go round: it warns everybody about the new policy on their old releases, but then the counter starts from 0. In other words they have 180 days, or however old the release is now (90 days or more) plus another 90 days. This added quite a lot complexity, but I think is worth doing.

There are all sorts of challenging race conditions and so on, and the net result is that releases should be deleted, so we should be careful to ensure that this works, and probably do a rollout only on dev. I'm publishing this to `altera` so that it _doesn't_ automatically deploy to the dev container, which comes from the `sbp` branch. As another safety measure, the actual deletions are protected behind a boolean. In other words, this PR won't actually delete anything anyway, but to do so all we have to do is change one flag.

There are lots of inelegant things about this PR that I dislike. One of the main ones is that I had to duplicate a lot of code to the foundation admin role in the storage interface, because the system caller needs it. Review tools _very strongly_ recommended adding yet another new class in the role hierarchy for system calls. I did add something similar to that just recently, allowing system calls to say "I have this ASF UID and I'm acting on behalf of it, but I don't want you to authenticate anything about that ASF UID as I vouch for it myself". That's fairly easy in comparison, however, because it uses the existing roles in the hierarchy. I've also just added participant release managers as a new role, which is the first new role since the storage interface was created, and am not keen on adding another one already.

### DIFF
--- a/atr/get/start.py
+++ b/atr/get/start.py
@@ -108,10 +108,6 @@ def _get_phase_symbol(phase: sql.ReleasePhase) -> str:
             return "Ⓡ"
 
 
-def _pre_submit(project: sql.Project) -> htm.Element | None:
-    return _cycle_preview(project)
-
-
 async def _render_page(project: sql.Project, releases: list[sql.Release]) -> htm.Element:
     page = htm.Block()
 
@@ -129,7 +125,7 @@ async def _render_page(project: sql.Project, releases: list[sql.Release]) -> htm
         submit_label="Start new release",
         cancel_url=util.as_url(root.index),
         defaults={"project_key": project.key},
-        pre_submit=_pre_submit(project),
+        pre_submit=_cycle_preview(project),
         skip=["auto_archive_prior"] if not project.policy_auto_archive_prior_release else [],
     )
     if releases:
--- a/atr/get/start.py
+++ b/atr/get/start.py
@@ -108,21 +108,8 @@ def _get_phase_symbol(phase: sql.ReleasePhase) -> str:
             return "Ⓡ"
 
 
-def _inactivity_policy_notice() -> htm.Element:
-    return htm.div(".alert.alert-light.mt-2.small")[
-        htm.strong["Inactivity policy: "],
-        "Release candidate drafts inactive for 90 days are cleaned up automatically. "
-        "A warning email is sent at 80 days.",
-    ]
-
-
-def _pre_submit(project: sql.Project) -> htm.Element:
-    children: list[htm.Element] = []
-    cycle = _cycle_preview(project)
-    if cycle is not None:
-        children.append(cycle)
-    children.append(_inactivity_policy_notice())
-    return htm.div[*children]
+def _pre_submit(project: sql.Project) -> htm.Element | None:
+    return _cycle_preview(project)
 
 
 async def _render_page(project: sql.Project, releases: list[sql.Release]) -> htm.Element:
--- a/atr/get/start.py
+++ b/atr/get/start.py
@@ -108,6 +108,23 @@ def _get_phase_symbol(phase: sql.ReleasePhase) -> str:
             return "Ⓡ"
 
 
+def _inactivity_policy_notice() -> htm.Element:
+    return htm.div(".alert.alert-light.mt-2.small")[
+        htm.strong["Inactivity policy: "],
+        "Release candidate drafts inactive for 90 days are cleaned up automatically. "
+        "A warning email is sent at 80 days.",
+    ]
+
+
+def _pre_submit(project: sql.Project) -> htm.Element:
+    children: list[htm.Element] = []
+    cycle = _cycle_preview(project)
+    if cycle is not None:
+        children.append(cycle)
+    children.append(_inactivity_policy_notice())
+    return htm.div[*children]
+
+
 async def _render_page(project: sql.Project, releases: list[sql.Release]) -> htm.Element:
     page = htm.Block()
 
@@ -125,7 +142,7 @@ async def _render_page(project: sql.Project, releases: list[sql.Release]) -> htm
         submit_label="Start new release",
         cancel_url=util.as_url(root.index),
         defaults={"project_key": project.key},
-        pre_submit=_cycle_preview(project),
+        pre_submit=_pre_submit(project),
         skip=["auto_archive_prior"] if not project.policy_auto_archive_prior_release else [],
     )
     if releases:
--- a/atr/storage/__init__.py
+++ b/atr/storage/__init__.py
@@ -27,6 +27,7 @@ from typing import TYPE_CHECKING, Any, Final
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator
 
+import atr.constants as constants
 import atr.db as db
 import atr.log as log
 import atr.models.basic as basic
@@ -37,8 +38,6 @@ import atr.storage.outcome as outcome
 import atr.storage.readers as readers
 import atr.storage.writers as writers
 import atr.user as user
-
-SYSTEM_ASF_UID: Final[str] = "system"
 
 # Access
 
@@ -600,7 +599,7 @@ async def write_as_project_release_manager(
 @contextlib.asynccontextmanager
 async def write_as_system() -> AsyncGenerator[writers.release.FoundationAdmin]:
     async with db.session() as data:
-        system_authorisation = types.SimpleNamespace(asf_uid=SYSTEM_ASF_UID)
+        system_authorisation = types.SimpleNamespace(asf_uid=constants.SYSTEM_SERVICE_UID)
         system_write: Any = types.SimpleNamespace(authorisation=system_authorisation)
         system_write_as: Any = WriteAs()
         yield writers.release.FoundationAdmin(system_write, system_write_as, data)
--- a/atr/storage/__init__.py
+++ b/atr/storage/__init__.py
@@ -21,7 +21,8 @@ import contextlib
 import datetime
 import json
 import logging
-from typing import TYPE_CHECKING, Final
+import types
+from typing import TYPE_CHECKING, Any, Final
 
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator
@@ -36,6 +37,8 @@ import atr.storage.outcome as outcome
 import atr.storage.readers as readers
 import atr.storage.writers as writers
 import atr.user as user
+
+SYSTEM_ASF_UID: Final[str] = "system"
 
 # Access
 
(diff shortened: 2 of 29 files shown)


sbp:
Oh, and we should centralise a constant for the system UID, actor, whatever we should call it. This is a general thing that we should clean up across the whole codebase one day, but it's low priority. (Duplicate strings that do the same thing should be symbolic constants.)

dave2wave:
> This PR sends out warnings if a release has been inactive for 80 days. When it reaches 90 days, it deletes it. It makes an exception on the first go round: it warns everybody about the new policy on their old releases, but then the counter starts from 0. In other words they have 180 days, or however old the release is now (90 days or more) plus another 90 days. This added quite a lot complexity, but I think is worth doing.

I think 180 days is way too long. I think that 60 day warning with a 30 day grace is sufficient.

sbp:
@alitheg Dave and I had a meeting about this, and one outcome is that we think that existing 90 day or older releases at the moment that this feature is deployed (only) should be treated as if they are 80 days old. There are various ways in which this could be done, but I guess it would be easiest during the Alembic backfill. That way, they get a warning but then are treated in the same way as new releases after the feature is deployed, which should simplify the code. They still have 10 days to prevent deletion. Another outcome is that there should be a button that lets people reset the clock to 0, instead of having to figure out which actions reset the clock. Also it would be great if we could use the notification mechanism to point people to the reset form, though that would require supporting HTML in the notifications.

sbp:
@alitheg I've updated it so that it rewrites all inactivity periods over 80 days to exactly 80 days during the backfill, so that the logic is much simpler. I've also added forms to the compose and finish pages allowing users to reset the inactivity clock without actually doing anything to the release. The vote page doesn't have one because when the vote ends, the inactivity will reset anyway, and starting a vote also would have reset inactivity.

atr/storage/writers/release.py
```diff
@@ -1170,3 +1254,205 @@ def __init__(self, write: storage.Write, write_as: storage.WriteAsFoundationAdmi
...
+
+    async def __check_eligibility(
+        self,
+        project_key: safe.ProjectKey,
+        version: safe.VersionKey,
+    ) -> str | None:
+        release = await self.__data.release(project_key=str(project_key), version=str(version), _committee=True).get()
+        if release is None:
+            return "release no longer exists"
+        if release.project.status != sql.ProjectStatus.ACTIVE:
```
alitheg, line 1556: I've been wondering if we should add a property onto project called "is_active" rather than repeating this check everywhere. What do you think?

alitheg: We could add one for "days_since_active" as well

atr/storage/__init__.py
```diff
@@ -37,6 +38,8 @@
 import atr.storage.writers as writers
 import atr.user as user
 
+SYSTEM_ASF_UID: Final[str] = "system"
```
alitheg: I had to move this to `constants.py` for the token work - because storage already import jwtoken so I couldn't do that import in reverse.

atr/user.py
```diff
@@ -106,6 +110,14 @@ def is_committer(committee: sql.Committee | None, uid: str) -> bool:
     return any((committer_uid == uid) for committer_uid in committee.committers)
 
 
+def is_participant_for_committee(committee: sql.Committee | None, project_keys: Iterable[str]) -> bool:
+    if committee is None:
+        return False
+    if config.is_test_mode() and (committee.key == "test"):
+        return True
+    return committee.key in project_keys
```
alitheg, line 118: Isn't this comparing a committee key to a list of project keys? It looks like it's projects being passed in at all the call sites

atr/storage/__init__.py
```diff
@@ -594,6 +597,15 @@ async def write_as_project_release_manager(
         yield await w.as_project_release_manager(project_key)
 
 
+@contextlib.asynccontextmanager
+async def write_as_system() -> AsyncGenerator[writers.release.FoundationAdmin]:
+    async with db.session() as data:
+        system_authorisation = types.SimpleNamespace(asf_uid=SYSTEM_ASF_UID)
```
alitheg: You mentioned the strong suggestion to add a new class - would that be along the lines I did in the system PAT work? Perhaps with two use cases it is necessary?

atr/storage/writers/release.py
```diff
@@ -157,15 +164,55 @@ def __init__(
...
+            .where(
+                via(sql.Release.key) == release.key,
+                via(sql.Release.phase).in_(_ACTIVITY_BUMP_PHASES),
+            )
+            .values(activity_at=activity_at_value, inactivity_notice_key=None)
+        )
+        if getattr(result, "rowcount", 0) != 1:
+            await self.__data.rollback()
+            raise storage.AccessError("The release state has changed, please refresh and try again", status=409)
+        await self.__data.refresh(release)
```
alitheg, line 209: Are you refreshing here to deal with a potential TOCTOU-style issue in its existence? If something has deleted the release before now I think this would raise and then the transaction wouldn't roll back?

atr/tasks/inactivity.py
```diff
@@ -0,0 +1,385 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+NOTICE_KIND_PREVIEW_ESCALATION: Final[str] = "preview_escalation"
+NOTICE_KIND_WARNING: Final[str] = "warning"
+
+_DELETION_ENABLED: Final[bool] = False
+_INACTIVITY_DELETE_DAYS: Final[int] = 90
+_INACTIVITY_WARNING_DAYS: Final[int] = 80
+_SYSTEM_ACTOR: Final[str] = "system"
```
alitheg: We should declare this only once, instead of here and in storage.

atr/get/start.py
```diff
@@ -108,6 +108,10 @@ def _get_phase_symbol(phase: sql.ReleasePhase) -> str:
             return "Ⓡ"
 
 
+def _pre_submit(project: sql.Project) -> htm.Element | None:
+    return _cycle_preview(project)
+
+
```
alitheg: Are you planning to add more here?

alitheg (changes requested):

atr/storage/writers/release.py
sbp: Sure, good idea. I'll do that in a commit separate to this PR.

atr/storage/__init__.py
sbp: In your PR I see `asf-system` used for this, but we run maintenance tasks under `system` and the inactivity task that I just added (and this PR too, to be consistent) uses that as well. Your PR doesn't change any of those existing uses, so I wonder if we should standardise on one or the other. If we move to `asf-system`, we'd have to consider whether that'll cause problems with any e.g. existing queued maintenance tasks.

sbp: I'm going to change them all to use the symbol, and set the symbol to `system` for now, in the same place and using the same constant name that you chose. If we decide to migrate to `asf-system` afterwards, I think we should do that as a separate PR so that we can consider the ramifications in isolation.

atr/user.py
sbp: No, these are committees. This is very confusing, has always been confusing, and we should decide what to do about it. The problem is that the naming here is aligned to _upstream_, which in this case is LDAP. We get the list of projects from `ou=project,ou=groups` and that means "committees that the user is a committer of". Those are literally called projects in LDAP, which is why we use the name, but really they are committees. For committee membership, it LDAP uses `pmc`. This causes confusion in the existing code, not just in this PR, for example in `atr/get/keys.py`:

```python
            project_list = session.committees + session.projects
            user_committees = await data.committee(name_in=project_list).all()
```

Here you can see we blend committees and projects, which looks like an error, but that's just committees of which the user is a member and committees of which the user is a committer. (The former should be a subset of the latter, but there are apparently very rare cases where that is not so.) So really this is just mixing two lists of committees. We should actually make a set out of this anyway, but maybe sqlite already does that. Anyway, it's confusing.

I think the best thing to do here would be to rename to ATR terminology as soon as we have control over the names ourselves. This should be in a commit separate to this PR.

atr/storage/__init__.py
sbp: Yeah, we should probably do that. Again, I think that should be separate to this PR. I'm saying that for almost all of your feedback, I realise, but this PR is already 1900 lines of additions and horrendously overcomplicated.

alitheg: Agreed!

atr/user.py
alitheg: We should probably at least call the param in a function like this "committee_keys" - try and keep the confusion isolated to just the session (so we'd still be passing in session.projects but at least this function would make semantic sense?

atr/storage/writers/release.py
sbp: https://github.com/apache/tooling-trusted-releases/issues/1282

atr/storage/__init__.py
alitheg: Ah yes I forgot I'd used `system` in the maintenance tasks. We can stick with that for now, agreed. Easy to do as a separate PR if we want to change it.

sbp: https://github.com/apache/tooling-trusted-releases/issues/1283

atr/user.py
sbp: https://github.com/apache/tooling-trusted-releases/issues/1284

atr/storage/__init__.py
sbp: https://github.com/apache/tooling-trusted-releases/issues/1285

atr/storage/writers/release.py
sbp: Nope, the rowcount check is the TOCTOU check. The refresh is to get the current `activity_at`. This should be serialising in the write lock, so it shouldn't raise, but even if it does then ultimately it's in the caller's `async with storage.write(...)`, which would handle the transaction and rollback properly anyway.

atr/tasks/inactivity.py
sbp: I'll replace `_SYSTEM_ACTOR` with `constants.SYSTEM_SERVICE_UID`. I'll put these:

```python
INACTIVITY_DELETE_DAYS: Final[int] = 90
INACTIVITY_WARNING_DAYS: Final[int] = 80
```

In `constants.py`.

atr/get/start.py
sbp: Heh, sorry, this was refactoring junk. Will fix this!

sbp:
@alitheg Thanks for the review! Resolved all of your comments either directly in the code in 8be81745665e6a95658ad424e45087a341d2d521 or by opening new issues (#1282, #1283, #1284, #1285).

sbp:
I've updated Puppet to use `tertia` instead of `sbp` as the source of the `atr-dev` container. I've copied this PRs commits into a single squashed commit in `tertia`, and set `_DELETION_ENABLED: Final[bool] = True` so that it will actually run on the dev server. There are only two releases on the dev server that will be deleted:

```
=== Releases that will be clamped and warned ===
2 releases

  tooling-0.1                                        RELEASE_CANDIDATE_DRAFT   2025-11-03 20:55:32.913228  (205d)
  tooling-0.2                                        RELEASE_CANDIDATE_DRAFT   2026-02-23 17:59:51.695972  (93d)
```

And I am the only person who will be notified. Ten days from today is Sunday, 7th June, so I should receive the messages and be able to check whether it worked on 8th June. If all goes well with the deployment to dev today, I will deploy the same code to `atr-staging` tomorrow. Those deletions will happen on 8th June.

In case of emergency, set `_DELETION_ENABLED: Final[bool] = False` and commit.

PR #1276, opened 2026-05-25
System PATs

alitheg:
Add a new "system" level PAT. Despite the name, they don't get access to everything, but they can be enabled for specific endpoints with a new auth decorator. PATs can be IP limited now (but the option to do so is only shown on the system PAT page which is an admin-only page). The use case for this is the .asf.yaml gitbox process which will need a long-lived credential. This will be 6 months from this PR but we can obviously revisit that in future.

### DIFF
--- a/atr/storage/__init__.py
+++ b/atr/storage/__init__.py
@@ -271,6 +271,25 @@ class WriteAsUserService(WriteAs):
         return self.__asf_uid
 
 
+class WriteAsSystemService(WriteAs):
+    # Built directly from the service identity rather than via Authorisation,
+    # which would reject an identity with no LDAP account.
+    def __init__(self, data: db.Session, asf_uid: str):
+        asf_uid = asf_uid.strip()
+        # Only the fixed service identity is valid here; anything else means
+        # a caller has wired this up wrongly.
+        if asf_uid != constants.SYSTEM_SERVICE_UID:
+            raise AccessError("System service writes require the system service identity", status=401)
+        self.__asf_uid = asf_uid
+        self.policy = writers.policy.SystemService(self, data)
+        self.project = writers.project.SystemService(self, data)
+        self.tokens = writers.tokens.SystemService(self, data)
+
+    @property
+    def asf_uid(self) -> str:
+        return self.__asf_uid
+
+
 # TODO: Could name this WriteDispatcher
 class Write:
     # Read and Write have authenticator methods which return access outcomes
@@ -509,6 +528,13 @@ def ensure_project_active(project: sql.Project) -> None:
     raise AccessError(f"Project '{project.key}' is archived; release actions are disabled.")
 
 
+async def pat_is_system(pat_text: str) -> bool:
+    token_hash = writers.tokens.hash_pat(pat_text)
+    async with db.session() as data:
+        pat = await data.personal_access_token(token_hash, is_system=db.NOT_SET).get()
+    return bool(pat and pat.is_system)
+
+
 @contextlib.asynccontextmanager
 async def read(asf_uid: principal.UID = principal.ArgumentNone) -> AsyncGenerator[Read]:
     if asf_uid is principal.ArgumentNone:
@@ -603,6 +629,13 @@ async def write_as_system() -> AsyncGenerator[writers.release.FoundationAdmin]:
         system_write: Any = types.SimpleNamespace(authorisation=system_authorisation)
         system_write_as: Any = WriteAs()
         yield writers.release.FoundationAdmin(system_write, system_write_as, data)
+
+
+@contextlib.asynccontextmanager
+async def write_as_system_service(asf_uid: str) -> AsyncGenerator[WriteAsSystemService]:
+    async with db.session() as data:
+        wss = WriteAsSystemService(data, asf_uid)
+        yield wss
 
 
 @contextlib.asynccontextmanager
--- a/atr/storage/readers/tokens.py
+++ b/atr/storage/readers/tokens.py
@@ -46,12 +46,9 @@ class FoundationCommitter(GeneralPublic):
         if asf_uid is None:
             raise ValueError("Not authorized")
         via = sql.validate_instrumented_attribute
-        stmt = (
-            sqlmodel.select(sql.PersonalAccessToken)
-            .where(sql.PersonalAccessToken.asfuid == asf_uid)
-            .order_by(via(sql.PersonalAccessToken.created))
+        tokens = await (
+            self.__data.personal_access_token(asfuid=asf_uid).order_by(via(sql.PersonalAccessToken.created)).all()
         )
-        tokens = await self.__data.query_all(stmt)
         return [types.PersonalAccessTokenSafe.from_sql(token) for token in tokens]
 
     async def most_recent_jwt_pat(self) -> types.PersonalAccessTokenSafe | None:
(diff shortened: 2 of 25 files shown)


atr/storage/readers/tokens.py
```diff
@@ -46,12 +46,9 @@ async def own_personal_access_tokens(self) -> list[types.PersonalAccessTokenSafe
         if asf_uid is None:
             raise ValueError("Not authorized")
         via = sql.validate_instrumented_attribute
-        stmt = (
-            sqlmodel.select(sql.PersonalAccessToken)
-            .where(sql.PersonalAccessToken.asfuid == asf_uid)
-            .order_by(via(sql.PersonalAccessToken.created))
+        tokens = await (
+            self.__data.personal_access_token(asfuid=asf_uid).order_by(via(sql.PersonalAccessToken.created)).all()
```
sbp, line 50: So now `asfuid` on PATs means two different things depending on whether `is_system` is set. It either means owner for `is_system` being `False`, or creator for `is_system` being `True`. I think it _might_ be better to let `asfuid` be nullable (it currently isn't) and then `None` indicates a system user, but then we'd need a new `creator` field to record who created it, and that isn't very elegant either.

Anyway, the current design is, I think, propagating fragility because e.g. this line appears now to be wrong since it doesn't have `is_system=False`. Or do we want to show system PATs that a user created as well as their own PATs? If the latter, though, then we'd need to distinguish them in the UI.

sbp (changes requested):

atr/storage/readers/tokens.py
alitheg: Isn't creator/owner the same? I intentionally made it set the user who created it as I thought that was the case for regular PATs. It's just that regular PATs authenticate you *as* that user whereas the system ones don't.

But yes you're right - system PATs should only show in the admin view - I could default is_system to False in `__data.personal_access_token`?

atr/storage/__init__.py
```diff
@@ -269,6 +269,23 @@ def asf_uid(self) -> str:
         return self.__asf_uid
 
 
+class WriteAsSystemService(WriteAs):
+    # Built directly from the service identity rather than via Authorisation,
+    # which would reject an identity with no LDAP account.
+    def __init__(self, data: db.Session, asf_uid: str):
+        asf_uid = asf_uid.strip()
+        if not asf_uid:
+            raise AccessError("System service writes require a service identity", status=500)
```
sbp: Does this actually check the service identity anywhere, though? If not, perhaps we should check that it's equal to `jwtoken.SYSTEM_SERVICE_UID`, but I think that this UID probably belongs in the storage interface itself, or some module that both `jwtoken` and the storage interface import. They both import `db`, but that's a strange place to put it. I think that importing the storage interface into `jwtoken` is okay though.

sbp (changes requested):
Another thing, sorry; didn't include this in the first round.

atr/storage/readers/tokens.py
sbp: The difference (which may only exist in my mind, and may not be practical) is that a non-system PAT for `asfuid=alice` means that the bearer of that PAT authenticates as `alice` and can do anything that Alice can do. A system PAT for `asfuid=alice` means that the bearer of that PAT authenticates as a system service, not as Alice, and can do anything that the system service, not Alice, can do. That seems like a fundamental intrinsic difference, completely separating what the PATs do.

alitheg: Yes I see what you mean - I suppose the way I was looking at it was that if `alice` creates a PAT with `is_system` then it can do one thing and if she creates one without `is_system` then it can do something else. It's still a PAT in that it must be exchanged for a JWT, and it's still *very* relevant who's token it is, if something goes wrong.

Easy to null it and have a creator field instead, if you prefer.

atr/storage/__init__.py
alitheg: Ah, you're right it would make sense here. It *is* gated, in that this interface is only invoked from an endpoint which is decorated to require a system JWT, and the exchange of a PAT for that JWT checks the identity. But it should be checked here too.

atr/storage/readers/tokens.py
sbp: Yeah, interesting, two very different ways of looking at the same thing. I'm not sure if null and a creator field is better, but I think that making up ASF UIDs like `asf_uid="atr-system"` is probably not a good idea. They're not really ASF UIDs! If we thought of them as _ATR_ UIDs then maybe the problem just goes away, but as it is we're at constant risk (however negligible) of collisions, and it muddies the divide between people and services. It's also very confusing, because we do `asf_uid="atr-system"` for the storage interface, but the model of the PAT will say `asfuid="alice"`. I don't have a solution, though.

atr/storage/__init__.py
sbp: Thanks. My intention for the storage interface was that it does all the authoritative gating, so anything that's done by callers is the extra part, rather than the storage interface being extra. (We can always change this design, of course.)

sbp: In this case, I guess that both are authoritative in their own domains! The route in token A/A, and the storage interface in authorisation.

atr/storage/readers/tokens.py
alitheg: Yes, I suppose the problem here is that by calling the param on the storage interface `asfuid` were saying we only ever expect a named person to be using it. Which a system token isn't. The storage interface could call it "identity" instead for example - purely semantic but could be a resolution for part of it.

alitheg: Or, we could say system access doesn't use the storage interface, but that doesn't feel right to me.

alitheg: I suppose another possible solution could be that for long-lived service tokens they still access on behalf of the user who created them, who should be an admin anyway, and use `is_system` just to gate access to that part of the storage interface. But it wouldn't really be correct in my mind because, for example, when gitbox runs an update to a project metadata it's not actually being called by "me" and it wouldn't be accurate to say that's who did the update.

sbp: Yeah, I think that latter part:

> But it wouldn't really be correct in my mind because, for example, when gitbox runs an update to a project metadata it's not actually being called by "me" and it wouldn't be accurate to say that's who did the update.

Is why I'm uncomfortable with the dual use of `asfuid` on the PAT too. When we call the storage interface, we don't mind who created the PAT, I think? We're only interested that it's a valid system PAT. I'm also less convinced, the more that I think about it, that we need to record the coiners of the system PATs in the system itself. Is that more akin to runbook information? When I want to revoke a system PAT, for example, do I have to spoof your user if you created that system PAT? It's more like a group ownership, I guess. The creations, and hence the metadata about who created such a PAT, should probably be audit log events only.

dave2wave:
A general comment on tracking the admin user for the PAT. What should the course of events when an Admin user who was issued a PAT is no longer an admin user?

If the PATs are limited to specific endpoints and several IP addresses then we should accept the PAT regardless of who created it, but if the PAT is useful even after the account is no longer qualified then we have do revocation, etcetera.

We should also provide warning emails when System PATs are nearing expiration to an address that Infra chooses.

sbp:
Because user role comes from LDAP, the PAT string and all of its associated metadata in ATR can remain unchanged when a user role changes (e.g. admin to not admin). For the system account, it shouldn't matter who originally requested the PAT, but when the requesting user role changes then we would need to take commensurate action.

atr/storage/readers/tokens.py
alitheg: Updated - I added a created column and allow asfuid to be null for the system ones.

Issue #1286, opened 2026-05-27
Document how to move release files out of the ATR SVN area


sbp:
So that publications can be complete.

dave2wave:
Closed by 4edfebd8a13ef2dc129a5ad0b38951228c7ff7a3

PR #1289, opened 2026-05-28
Document how to move files in SVN dist for publication

sbp:
I didn't verify that the commands work as my user because I don't have the necessary permissions. I could use the token, but then it wouldn't be testing this workflow. Also I don't want to move test files into `dist/release` anyway, for obvious reasons.


sbp:
This is for #1286.

atr/docs/promoting-to-release.md
```diff
@@ -0,0 +1,109 @@
...
+* [Overview](#overview)
+* [Prerequisites](#prerequisites)
+* [Moving the files](#moving-the-files)
+* [Verifying the move](#verifying-the-move)
+* [Removing superseded releases](#removing-superseded-releases)
+* [The KEYS file](#the-keys-file)
+
+## Overview
+
+When the ATR workflow completes, i.e. when the vote is resolved and the release announced, ATR commits the approved artifacts to its own area of the Apache distribution SVN repository:
```
dave2wave, line 22: Please add a note with a > section that this is Alpha test behavior which will change for the Beta

atr/docs/promoting-to-release.md
```diff
@@ -0,0 +1,109 @@
...
+```shell
+svn rm -m "Remove superseded <committee> <previous-version>" \
+  https://dist.apache.org/repos/dist/release/<committee>/<previous-version>
+```
+
+Do this as a separate commit, after promotion, once you have verified that the new release is reachable.
+
+## The KEYS file
+
+The `KEYS` file lives at `https://dist.apache.org/repos/dist/release/<committee>/KEYS` and is managed independently of any individual release. ATR does not publish it. If a new signer's key is used for this release, update `KEYS` from your local SVN checkout in a separate commit before users begin verifying artifacts.
```
dave2wave: There needs to be more detail here about where ATR has written the new KEYS file. Do we have a gap in how we are discussing the KEYS file for Alpha?

dave2wave (changes requested):

atr/docs/promoting-to-release.md
sbp: Yeah, we import existing `KEYS` files and then export them (in a different style to the imported files) currently on ATR only. We don't publish them to SVN. Perhaps we should do that? One safety check we could do is to show a diff of what keys have been added compared to the `KEYS` file that is currently in SVN.

sbp:
@dave2wave Updated in 552d3a215c3d637b1dd268614342087ac88b3f9f.

atr/docs/promoting-to-release.md
```diff
@@ -107,3 +109,5 @@ Do this as a separate commit, after promotion, once you have verified that the n
 ## The KEYS file
 
 The `KEYS` file lives at `https://dist.apache.org/repos/dist/release/<committee>/KEYS` and is managed independently of any individual release. ATR does not publish it. If a new signer's key is used for this release, update `KEYS` from your local SVN checkout in a separate commit before users begin verifying artifacts.
+
+ATR imports all keys from existing `KEYS` files automatically, if it can, and allows users to associate additional new keys with committees. These `KEYS` files are then exported to ATR's own download area in Alpha, but this download area will be removed when SVN publication is out of testing. We are currently discussing the mechanisms for exporting and managing `KEYS` files in ATR going forward.
```
dave2wave: Let's finish this discussion, please describe how to diff and update from the ATR generated KEYS file and to svn dist release

dave2wave (changes requested):

atr/docs/promoting-to-release.md
dave2wave: Perhaps we write KEYS files to our test svn area and commit.

atr/docs/promoting-to-release.md
dave2wave: Yes, please.

atr/docs/promoting-to-release.md
sbp: The diff would have to be a conceptual diff, created by OpenPGP aware code. The order of keys and the formatting of keys differs in the ATR serialised version. This code would be custom, and would probably work best as a tool accessed within ATR itself.

dave2wave: Let's put it this way the update should be safe as the order doesn't matter and this is an Alpha only stopgap.

dave2wave: Let's not spend time creating a diff and instead make KEYS update look like svn dist update and know that the file is safe.

sbp:
@dave2wave Updated and ready for a new review. Commit 2d58d3567179fb743ed57c8c3e63e02d4a604994 contains the changes from the previous review.

dave2wave (approved):
LGTM

PR #1376, opened 2026-07-08
#1370 - support import and export of PMC CSVs

### LINKED ISSUE #1370: Admin page for seeding Projects and Artifact Catalog
As we move to and support the beta we are going to need to accommodate fixing mistakes in creating subprojects and assigning artifacts. Some projects have a complex history. We'll need a page that will both:

1. Correct the subprojects for a PMC
2. Provide a seeding of Artifacts that assigns these to the correct project and release version.

alitheg:
Adds support for import/export of a PMC's projects, releases, artifacts by an admin.

For review - it's a decent sized change and likely wants some manual testing on dev once deployed! I can do that once it's merged to `tertia`

### DIFF
--- a/atr/admin/__init__.py
+++ b/atr/admin/__init__.py
@@ -17,6 +17,7 @@
 
 import asyncio
 import collections
+import dataclasses
 import datetime
 import hashlib
 import ipaddress
@@ -36,6 +37,7 @@ import htpy
 import jwt
 import pydantic
 import quart
+import quart.datastructures as datastructures
 import sqlalchemy
 import sqlmodel
 import werkzeug.exceptions as exceptions
@@ -61,6 +63,9 @@ import atr.noisy as noisy
 import atr.paths as paths
 import atr.principal as principal
 import atr.shared as shared
+import atr.shared.catalogue_diff as catalogue_diff
+import atr.shared.catalogue_import as catalogue_import
+import atr.shared.catalogue_rows as catalogue_rows
 import atr.storage as storage
 import atr.storage.outcome as outcome
 import atr.tasks as tasks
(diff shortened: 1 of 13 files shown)


atr/storage/writers/catalogue.py
```diff
@@ -227,6 +222,111 @@ async def delete_project(self, project_key: safe.ProjectKey) -> None:
...
+            moved=counts["release_move"],
+            rehomed=counts["artifact_rehome"],
+            conflicts=counts["conflict"],
+        )
+
+    async def _move_one_release(self, move: catalogue_diff.ReleaseMove) -> None:
+        via = sql.validate_instrumented_attribute
+        old_key = move.key
+        version = str(move.row.version)
+        new_key = f"{move.to_project}-{version}"
```
sbp: If the uploader gets the version wrong, couldn't this corrupt the release? There doesn't appear to be any validation of this `new_key`. I tested, and if there are artifacts then you get a 500 error anyway, but it breaks the whole import. If there are no artifacts, it corrupts silently. I assume that this will cause problems elsewhere, but I didn't check (I mean, release key is `proj-0.1` but version is `0.2`).

atr/shared/catalogue_diff.py
```diff
@@ -0,0 +1,228 @@
...
+    existing_pks = {(pk[0], pk[1], pk[2]) for pk in snapshot.artifact_by_dist.values()}
+    for row in rows:
+        reference_conflict = _artifact_reference_conflict(row, snapshot, known_projects, known_releases)
+        if reference_conflict is not None:
+            diff.conflicts.append(reference_conflict)
+            continue
+        path = str(row.artifact_path)
+        pk = (str(row.project_key), str(row.version), path)
+        suffix = str(row.download_path_suffix) if row.download_path_suffix else ""
+        matched = snapshot.artifact_by_dist.get((suffix, path)) if suffix else None
```
sbp, line 195: So this is looking up artifacts by `(download path, filename)`? Problem is, since download path is an exported field, it's also editable. If anybody changes it, how will it know what's being moved? So, any download path edit becomes a crash: it's an actual 500.

atr/shared/catalogue_diff.py
```diff
@@ -0,0 +1,228 @@
...
+            continue
+        target_committee = snapshot.project_committee.get(target_project) or pending.get(target_project)
+        if target_committee is None:
+            diff.conflicts.append(
+                Conflict(table="releases", key=key, reason=f"target project '{target_project}' does not exist")
+            )
+            continue
+        current_project = snapshot.release_project.get(key)
+        if current_project is None:
+            diff.adds.append(row)
```
sbp, line 267: I guess `if target_release_key in snapshot.release_keys` was intended to be the global check here (it's below in the code), but because there's a `continue` it's never reached. Basically this check is only looking at whether it's a release key in the existing committee, whereas the later check, which is not reached, would check whether it's a release key in the whole database. Small point, but also easy to fix.

atr/storage/writers/catalogue.py
```diff
@@ -227,6 +222,111 @@ async def delete_project(self, project_key: safe.ProjectKey) -> None:
...
+            .where(via(sql.Artifact.version) == version)
+            .values(release_key=new_key, project_key=move.to_project)
+        )
+        await self.__data.execute(
+            sqlmodel.update(sql.LifecycleEvent)
+            .where(via(sql.LifecycleEvent.version_key) == old_key)
+            .values(version_key=new_key, project_key=move.to_project)
+        )
+        # Other release-key children re-point to the new key
+        for model, attr in repoint.RELEASE_KEY_REFS:
```
sbp, line 403: This contains `sql.Task.version_key`, but it shouldn't do because that's not a release key (`example-0.1`), it's just the version (`0.1`):

```
RELEASE_KEY_REFS: Final[list[tuple[type[sqlmodel.SQLModel], str]]] = [
    (sql.Release, "key"),
    (sql.Artifact, "release_key"),
    (sql.LifecycleEvent, "version_key"),
    (sql.CheckResult, "release_key"),
    (sql.BallotPaper, "release_key"),
    (sql.Distribution, "release_key"),
    (sql.Quarantined, "release_key"),
    (sql.Revision, "release_key"),
    (sql.Task, "version_key"),
]
```

atr/shared/catalogue_diff.py
```diff
@@ -0,0 +1,228 @@
...
+    }
+    known_releases = (
+        set(snapshot.release_project)
+        | {str(add.key) for add in diff.adds if isinstance(add, catalogue_rows.ReleaseRow)}
+        | {f"{move.to_project}-{move.row.version}" for move in diff.release_moves}
+    )
+    existing_pks = {(pk[0], pk[1], pk[2]) for pk in snapshot.artifact_by_dist.values()}
+    for row in rows:
+        reference_conflict = _artifact_reference_conflict(row, snapshot, known_projects, known_releases)
+        if reference_conflict is not None:
```
sbp, line 185: This only seems to check the destination, not the source.

sbp: Below, `if current_project in snapshot.managed_project_keys` checks the source for releases. Artifact sources aren't checked.

sbp (changes requested):

atr/storage/writers/catalogue.py
sbp: So I think Tasks will have to be handled separately.

sbp: The `(sql.LifecycleEvent, "version_key")` entry looks suspicious too, but I didn't check it.

alitheg: Two issues here - 1. Claude was overzealous in finding release keys and I missed it (it identified version_key there because it *is* called that in LifecycleEvent - I'll raise a bug and fix later). 2. I neglected to note in this part that tasks don't need moved since the whole admin page doesn't let you edit projects that have already started using ATR properly (since mangling releases, artifacts etc including on-disk and draft etc. just seems fraught with peril). This is purely for catalogued projects/releases/artifacts which exist in SVN but have no ATR-managed files anywhere. Which won't have tasks!

atr/storage/writers/catalogue.py
alitheg: Quite right, and that hits something that bugged me when I did this - I've changed to start deriving the release keys instead of exposing them. For the scope of this page, it doesn't really matter that you can't _change_ a key, you create a new one and reassign the artifacts to it - you can always remove the empty release as a separate operation by merging it with another. 

There was another bug here - if you change version and nothing else your edit was silently dropped.

atr/shared/catalogue_diff.py
alitheg: The path _is_ the identity, so we should be using it to look up the existing one, and it's the only real conflict in seeding/correcting artifacts (you can't use the same full artifact path in more than one release). The add path now checks to make sure there's no conflict there, but I'm considering actually having the uploaded artifact CSV be an exhaustive list - remove all that don't match from the same PMC. This is for correcting the seeds, after all.

atr/shared/catalogue_diff.py
alitheg: Good spot. It's the two separate branches  - the latter being one for moving to another committee. So we don't actually check existence when we're adding a new row! That second check is still correct for moving a release, but we do need to check for conflict in the first case.

atr/shared/catalogue_diff.py
alitheg: This is a check actually intended to prevent using these tools to manipulate ATR-managed releases (where we'll have on-disk files, revisions, etc).  But we should be checking the source as well.

alitheg:
@sbp I've updated this so you can now choose to do an overwrite all. In most real cases, we'll be doing this - we'll exclude a complex PMC the seed script, then manually seed them through this. In some other cases we might want to just upload a few, which the additive path will now do. Additive is default in the UI since it's non-destructive. I need to do some testing, but further comments now welcomed/invited!

atr/admin/__init__.py
```diff
@@ -680,6 +685,228 @@ async def catalog_remove_post(
...
 
 
+_CATALOG_PREVIEW_LIMIT: Final[int] = 50
+
+
+class CatalogImportForm(form.Form):
+    projects: form.File = form.label("Projects CSV", "Optional.")
+    releases: form.File = form.label("Releases CSV", "Optional.")
+    artifacts: form.File = form.label("Artifacts CSV", "Optional.")
+    additive: bool = form.label(
```
sbp: This field is `default=True`, but because browsers don't send anything when the field is unchecked, that's when we apply the default. So when the field is checked, the browser sends `"on"` and we set `True`. If the field is unchecked, the browser sends nothing, and we use the default, which is `True`. Therefore it's impossible to turn this field off.

I wonder if we can add a static lint for this. Might be difficult because the field element is a consequence of the Python type of the property.

atr/shared/catalogue_diff.py
```diff
@@ -0,0 +1,494 @@
...
+        str(add.key) for add in diff.adds if isinstance(add, catalogue_rows.ReleaseRow)
+    }
+    seen_dists: set[tuple[str, str]] = set()
+    seen_pks: set[tuple[str, str, str]] = set()
+    for row in rows:
+        path = str(row.artifact_path)
+        pk = (str(row.project_key), str(row.version), path)
+        # Two rows sharing a dist path would be two catalogue entries for one file in svn
+        dist = (str(row.download_path_suffix), path) if row.download_path_suffix else None
+        if (pk in seen_pks) or ((dist is not None) and (dist in seen_dists)):
```
sbp: So `seen_dists` is what was seen in the file, not in the db? But ATR managed projects are skipped, so if the uploaded CSV contained a file in an unmanaged project that was actually in a managed project, would you not end up with a duplicate? We check that it's not duplicated within the CSV, that the target project isn't managed in ATR, that the committee and release exist, and that the primary key is free, but all of those would pass in this circumstance and yet there'd still be a duplicate.

atr/shared/catalogue_diff.py
```diff
@@ -0,0 +1,494 @@
...
+    # An artifact stays within its own committee; a cross-committee correction moves the whole
+    # release, which carries its artifacts
+    known_projects = _known_projects(snapshot, diff, committee_key)
+    known_releases = (
+        set(snapshot.release_project)
+        | {str(add.key) for add in diff.adds if isinstance(add, catalogue_rows.ReleaseRow)}
+        | {f"{move.to_project}-{move.row.version}" for move in diff.release_moves}
+    )
+    seen_dists: set[tuple[str, str]] = set()
+    seen_pks: set[tuple[str, str, str]] = set()
```
sbp, line 180: You can move a release in `releases`, and that moves its artifacts along with it. If you then do an add in `artifacts` (the separate field for that), using the same name as an artifact's destination in the previous step, the code creates two duplicate rows for the artifact. This then fails because of the unique constraint in the database.

Here's a test case for this one:

```
def test_classify_collision() -> None:
    snap = _snapshot(
        project_committee={"alpha-one": "alpha", "alpha-two": "alpha"},
        artifact_pks=frozenset({("alpha-one", "1.0.0", "a.tgz")}),
    )
    rows = {
        "releases": [_release("alpha-one-1.0.0", "alpha-two", "1.0.0")],
        "artifacts": [_artifact("alpha-two", "1.0.0", "a.tgz", "alpha/new")],
    }
    d = catalogue_diff.classify_additive(rows, snap, "alpha")
    assert d.counts["conflict"] == 1
```

(This fails, but it shouldn't.)

sbp (commented):
Not finished yet. May have a few more findings to come in a subsequent review, but you could commit if you like and we can do those as fixups.

atr/admin/__init__.py
alitheg: Ah, valid. Yes we should consider a lint for it as a possibility. I've reversed the logic so it's now checked == overwrite

atr/shared/catalogue_diff.py
alitheg: I've unified the logic between the two paths and made sure we're checking existing managed projects too, not just the snapshot of projects which match those uploaded.

atr/shared/catalogue_diff.py
alitheg: Thanks for the test, and good catch! It now tracks the resulting primary keys as it goes along, meaning we should catch duplicates that are both in the uploaded files.

atr/shared/catalogue_import.py
```diff
@@ -0,0 +1,312 @@
...
+        "sbom_path",
+        "classification",
+        "download_path_suffix",
+        "managed",
+        "dated",
+    ),
+}
+
+# The uploaded CSVs live here, one directory per token
+IMPORT_ROOT: Final[pathlib.Path] = pathlib.Path(config.get().STATE_DIR) / "temporary" / "catalogue-imports"
```
sbp, line 59: Or `paths.get_tmp_dir` gives a `safe.StatePath`.

sbp (commented):
One tiny comment. I was going to comment on the round tripping breaking, but that appears to be fixed now!

alitheg:
Closing as I'll merge manually and don't want GH to delete my branch :)

PR #1394, opened 2026-07-10
Adjust downloads menu target

### LINKED ISSUE #1393: Remove the downloads area feature
When ATR goes to beta, we will no longer support ATR's own downloads area, as this functionality will be entirely taken over by SVN. Therefore the downloads area must be removed before going to beta.

dave2wave:
Along with P6 changes fixes #1393

### DIFF
--- a/atr/templates/includes/topnav.html
+++ b/atr/templates/includes/topnav.html
@@ -93,13 +93,8 @@
                 Committees</a>
             </li>
             <li>
-              {% if is_test_mode %}
-                <a class="dropdown-item" href="/published/"><i class="bi bi-download"></i> Downloads</a>
-              {% else %}
-		{# This path is handled by the frontend proxy server #}
-		{# https://djlint.com/docs/ignoring-code/ #}
-		{# djlint:off J018 #} <a class="dropdown-item" href="/downloads/"><i class="bi bi-download"></i> Downloads</a> {# djlint:on #}
-              {% endif %}
+              <a class="dropdown-item" href="https:///downloads.apache.org/"><i class="bi bi-download"></i>
+                Downloads</a>
             </li>
           </ul>
         </li>


dave2wave:
Before merging the following need to be handled:

In `path.py`:

```
def committee_downloads_url(host: str, committee: sql.Committee) -> str:
    # This is a slight extension of the intended paths concept                                                                                                             
    # But URLs contain paths, so atr.paths does not have to be limited to filesystem paths                                                                                 
    if committee.is_podling:
        return f"https://{host}/downloads/incubator/{committee.key}"
    return f"https://{host}/downloads/{committee.key}"
```

And in `util.py`:

```
def download_url_for_path(
    relpath: safe.RelPath, kind: DownloadFile, host: str | None = None, archived: bool = False
) -> str:
    # Where a file at a known dist-relative path is fetched from, by lifecycle (archived vs beta vs                                                                        
    # released) and role (artifact vs metadata). The catalogue stores each artifact's exact path, so                                                                       
    # nothing has to be reconstructed from a committee here                                                                                                                
    if archived:
        # archive.apache.org keeps every release ever published, on every host                                                                                             
        return paths.archive_download_url(relpath)
    if not config.get().SVN_PUBLISH_URL:
        # No SVN publishing configured (dev), so ATR serves the downloads itself                                                                                           
        return f"https://{host or config.get().APP_HOST}/downloads/{relpath}"
    match svn_publish_target():
        case SvnPublishTarget.ATR:
            # Beta: files are still in ATR's area of the distribution SVN, not yet mirrored                                                                                
            return f"{config.get().SVN_DIST_PUBLIC_URL.rstrip('/')}/{relpath}"
        case SvnPublishTarget.RELEASE:
            if kind is DownloadFile.ARTIFACT:
                return paths.closer_download_url(relpath)
            return paths.downloads_url(relpath)
```

and in `api/__init__.py`

```
def _committee_keys_url(host: str, committee: sql.Committee) -> str:
    if committee.is_podling:
        return f"https://{host}/downloads/incubator/{committee.key}/KEYS"
    return f"https://{host}/downloads/{committee.key}/KEYS"
```

alitheg:
LGTM as far as the change itself. As far as I can see nothing calls `committee_downloads_url` or `_committee_keys_url` - and while `download_url_for_path` does need changed, it only serves `host/downloads` when SVN is unconfigured (so only ever in the dev environment)

dave2wave:
@sbp Since nothing calls `committee_downloads_url` or `_committee_keys_url` should we remove those functions?

sbp:
Isn't `committee_downloads_url` called by `util.public_download_url`? And then `util.public_download_url` has several callers throughout the codebase.

sbp:
Ah, perhaps these ones were intended:

```
def _committee_keys_path(committee: sql.Committee) -> safe.StatePath:
    downloads_dir = paths.get_downloads_dir()
    if committee.is_podling:
        return downloads_dir / "incubator" / committee.key / "KEYS"
    return downloads_dir / committee.key / "KEYS"


def _committee_keys_url(host: str, committee: sql.Committee) -> str:
    if committee.is_podling:
        return f"https://{host}/downloads/incubator/{committee.key}/KEYS"
    return f"https://{host}/downloads/{committee.key}/KEYS"
```

These two have no callers, and @alitheg added them in 8df94595a652155aca65dbb1ae38cc0fb91ca949. Okay to remove them, @alitheg?

alitheg:
Agreed for those two - they're redundant now. As for the others - looks like something is wrong with my "Find" function as it's also claiming `public_download_url` doesn't exist in the codebase... good job we spotted this!

dave2wave:
Since those routines are of no concern I will squash and merge

Issue #1400, opened 2026-07-13
Reclassify SBOMs as artifacts and add new check


alitheg:
Update the classifier to add a new class of file "SBOM" and classify SBOMs as that.

Add a new "suggestion" check which notes if a non-SBOM artifact does not have a matching SBOM artifact and provides instructions to create and sign one using the client.

sbp:
@alitheg I assume you're working on this one because of the PR, but if you want me to add the check please let me know!

alitheg:
Feel free to, depending on what else you're working on. I felt I could do
the classification quickly while waiting on a build for the other stuff.
Equally, I happy to take the rest of it.

alitheg:
Check implemented and showing on draft releases now. SBOMS classified properly

PR #1411, opened 2026-07-15
#1400 - Initial stab at SBOM check

alitheg:
@sbp Posting for your review since you've done more work on checks and classification in the past. This seems to work, although it has an issue in the real world: I have a release with a jar and a pom, both classify as bin, and both pass `analysis.is_artifact` so the checker wants two SBOMs.

### DIFF
--- a/atr/tasks/__init__.py
+++ b/atr/tasks/__init__.py
@@ -28,6 +28,7 @@ import atr.analysis as analysis
 import atr.attestable as attestable
 import atr.constants as constants
 import atr.db as db
+import atr.detection as detection
 import atr.hashes as hashes
 import atr.log as log
 import atr.models.args as args
@@ -43,6 +44,7 @@ import atr.tasks.checks.license as license
 import atr.tasks.checks.parity as parity
 import atr.tasks.checks.paths as paths
 import atr.tasks.checks.rat as rat
+import atr.tasks.checks.sbom as sbom_check
 import atr.tasks.checks.signature as signature
 import atr.tasks.checks.targz as targz
 import atr.tasks.checks.zipformat as zipformat
@@ -353,6 +355,8 @@ def resolve(task_type: sql.TaskType) -> Callable[..., Awaitable[results.Results 
             return distribution.status_check
         case sql.TaskType.DISTRIBUTION_WORKFLOW:
             return gha.trigger_workflow
+        case sql.TaskType.HAS_SBOM:
+            return sbom_check.check
         case sql.TaskType.HASHING_CHECK:
             return hashing.check
         case sql.TaskType.KEYS_IMPORT_FILE:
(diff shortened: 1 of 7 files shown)


sbp:
Sure thing, reviewing.

atr/tasks/__init__.py
```diff
@@ -714,45 +771,42 @@ async def _draft_file_checks(
...
+    if not analysis.is_artifact(path.as_path()):
+        return None
+    path_str = str(path)
+    return await queued(
+        asf_uid,
+        sql.TaskType.HAS_SBOM,
+        release,
+        revision_number,
+        path_str,
+        check_cache_key=await checks.resolve_cache_key(
```
sbp, line 802: This ends up only putting the artifact hash in the cache key, which means that if somebody is missing an SBOM, is notified by this check, and then uploads an SBOM, the new SBOM won't appear in the cache key so it won't be rechecked.

We could use the existing `all_files` resolver here, but I guess the proper thing to do would be to create a new resolver just for the candidate SBOM files? That's not particularly elegant either way, but I think the latter would be best. What do you think?

atr/tasks/__init__.py
```diff
@@ -714,45 +771,42 @@ async def _draft_file_checks(
...
+            resolve(sql.TaskType.HAS_SBOM),
+            sbom_check.CHECK_VERSION,
+            parity.INPUT_POLICY_KEYS,
```
sbp: Presumably should be `sbom_check.INPUT_POLICY_KEYS` (though it makes no difference currently since they're both `[]`).

atr/util.py
```diff
@@ -444,7 +444,7 @@ async def check_propagation(
 
 
 def checker_display_name(checker: str) -> str:
-    return checker.removeprefix("atr.tasks.checks.").replace("_", " ").replace(".", " ").title()
+    return checker.removeprefix("atr.tasks.checks.").replace("_", " ").replace(".", " ").title().replace("Sbom", "SBOM")
```
sbp, line 447: This only happens on compose pages, not on report pages, so it'll be Sbom on there. Not particularly high priority, but thought you might appreciate the note since I guess the intention is for this to be consistent.

sbp (commented):
On the main question about everything being told to have an SBOM, perhaps we could use `util.archive_format_stem` to limit the check to just "plain archive" artifacts?

atr/tasks/__init__.py
alitheg: I've created a resolver for file+suffixes. So we can cache on existence of any set of suffixes of any base file

alitheg:
> On the main question about everything being told to have an SBOM, perhaps we could use `util.archive_format_stem` to limit the check to just "plain archive" artifacts?

I just did that but it skips jars. Should ARCHIVE_FORMAT_SUFFIXES in that function be the same as QUARANTINE_ARCHIVE_SUFFIXES in `detection.py`?

sbp:
> Should ARCHIVE_FORMAT_SUFFIXES in that function be the same as QUARANTINE_ARCHIVE_SUFFIXES in `detection.py`?

Nah, they mean different things. Using `detection.QUARANTINE_ARCHIVE_SUFFIXES` as the suffixes of what to check might make sense though?

PR #1415, opened 2026-07-15
Remove links to ATR's own downloads area

sbp:
@alitheg This makes announcement emails use `downloads.apache.org` instead of `/downloads/` and changes the catalogue to use the same when it would have used `/downloads/`. One consequence is that in local development we can't really browse what has been published, because it's now expecting things to be published to SVN. We are still moving the finished files to `/finished/`, but it's not exposed, and I think in a subsequent change I might change that to just be attestable data instead. What do you think? And are there any problems with this PR? It affects the catalogue.

### DIFF
--- a/atr/construct.py
+++ b/atr/construct.py
@@ -199,8 +199,6 @@ async def announce_release_default(project_key: safe.ProjectKey) -> str:
 async def announce_release_subject_and_body(
     subject: str, body: str, options: AnnounceReleaseOptions
 ) -> tuple[str, str]:
-    host = config.get().APP_HOST
-
     async with db.session() as data:
         release = await data.release(
             project_key=str(options.project_key),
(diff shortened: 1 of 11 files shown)


dave2wave:
@sbp Technically announcement emails should use the now required downloads url.

alitheg:
Regarding local development, when I was testing the SVN work I created a local SVN repo which I was able to publish to. We could get that committed so that we can just use a local repo in local development.

sbp:
@dave2wave Ahh, yes! Thank you.

@alitheg Oh, I was thinking about that too. Glad you did the proof of concept already. Okay, let's make that part of how ATR works for local development indeed. That'll help to minimise the divergence between local dev and prod too, which is always good.

alitheg:
I *might* have not reset it all, I'll check tomorrow and get it running on
a bench if I have it

alitheg:
Ah actually @sbp it's already committed - `make svn-dev-repo` creates one in the state directory and gives you the export commands to support it

Issue #1417, opened 2026-07-16
Expired keys accepted for signing check


alitheg:
<div class="markdown-heading" dir="auto"><h4 tabindex="-1" class="heading-element" dir="auto">Expired signing keys accepted during signature verification; OpenPGP revocation handling not explicit or tested</h4></div>
<markdown-accessiblity-table data-catalyst="">
Attribute | Value
-- | --
Severity | 🔴 Critical
ASVS Level(s) | L2, L3
CWE | —
ASVS Sections | 11.6.1, 4.1.5
Files | atr/tasks/checks/signature.py, atr/storage/writers/keys.py, atr/get/keys.py
Source Reports | 11.6.1.md, 4.1.5.md
Related | —

</markdown-accessiblity-table>
<p dir="auto"><strong>Description:</strong></p>
<p dir="auto">ATR computes and persists key expiration as a first-class security attribute (<code>_key_expires_at</code> via <code>_effective_key_expiration_self_signature</code> in <code>atr/storage/writers/keys.py</code>, stored in <code>PublicSigningKey.expires</code>; <code>atr/get/keys.py::details</code> renders an 'Expired' danger badge from this field). However, the signature-verification path (<code>atr/tasks/checks/signature.py::_check_core_logic</code>) loads full <code>PublicSigningKey</code> rows — including <code>expires</code> — and never consults it. The <code>where</code> clause filters only on committee membership and <code>deleted IS NULL</code>; there is no <code>expires</code> filter, and <code>_check_core_logic_verify_signature</code> calls <code>signature.verify_file(candidate_key, artifact_path)</code>
 with no expiry check on the key or on the signature creation time 
relative to key validity. An expired key — precisely the class of keys 
most likely to have lapsed operational protection or been deliberately 
expired after suspected compromise — verifies identically to a valid 
key, yielding 'Signature verified successfully'. Data flow: 
attacker-held expired key (committee-linked, not soft-deleted) → 
detached <code>.asc</code> uploaded with tampered artifact → <code>check()</code> → <code>_check_core_logic</code> (no expiry filter) → <code>verify_file</code>
 succeeds. Impact: forged signatures on release artifacts pass the 
committee signature check, so the primary supply-chain assurance ('this 
artifact was signed by a currently valid committee key') is false. 
[...]

Issue #1419, opened 2026-07-16
Review the handling and treatment of signing subkeys throughout ATR


sbp:
We reference signatures to the primary key, but we measured that about 8% of releases so far use at least one subkey to sign, and two projects appear to use only subkeys. We need to review that we are treating subkeys correctly, exposing them throughout the UI appropriately, and that we have no related bugs.

This is associated with #1417.

alitheg:
I'm resolving this as Subkeys are now processed and stored in the database - opened #1439

PR #1422, opened 2026-07-17
#1417 - Support for subkeys

alitheg:
@sbp I think this handles subkeys reasonably - but there's a gap we should talk about: the keys page shows expiry from the database which is only expiry of the main key. It feels like a lot to log all the subkeys in the database, so I'm not sure the best way to make sure we can show signing key expiry to people

### DIFF
--- a/atr/pgp.py
+++ b/atr/pgp.py
@@ -1,0 +1,199 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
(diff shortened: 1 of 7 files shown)


alitheg:
Although I guess maybe that's #1419?

sbp (approved):
Great. I hadn't thought of `not_signing_key`.

sbp:
@alitheg Oh, and yeah, I think parsing out all the subkeys will be #1419. I know that'll create quite a few extra rows, but I expect that'll be the best solution when we do the research. It would be nice to show the composition of keys in ATR too, but that should probably be extra and during beta I'd have thought.

alitheg:
@sbp I've just pushed revocation handling to the `arm` branch if you'd like a look? I also had to change the cache key for the signature check as it was only keyed on fingerprint which was stable across the revocation, and needed to change the way we handled uploading a key that already existed  - because it was a soft delete, the key was just silently restored as before when I tried to upload a revoked one.

Issue #1450, opened 2026-07-27
Allow more tasks to be retried


sbp:
Checks record `exception` check results upon certain kinds of error such as network failures, but still return a task `success`. A finding of #1395 was that we want users to be able to restart checks that fail due to transient errors, but this strange hybrid status means that we cannot.

The idea of dual statuses was that if we catch exceptions, we knew that one might happen and the task handled it (as a check result) and therefore the task can continue to successful completion. Perhaps, instead, we should have two different kinds of task exceptions which indicate whether the task could be sensibly restarted or not.

sbp:
~I've decided to hold this feature as it is now less applicable to the later work than originally intended.~

sbp:
Whoops. Helps if you close the PR and not the issue!

sbp:
Resolved by 68169dbf8e8987bcaf7e55c11193b7e74b4fb694, 70f821b1f0f1bae73ecc6dfecf9f29db940f8cb9, 5e6bc5b28fb8b5510983aae96bb8565742dc5542, and 2d9ebc44fde1ae62d5e102cfaf079da2ed2469d0.

PR #1451, opened 2026-07-27
Prevent stale task workers from overwriting newer task results

sbp:
This is to prepare for further task work for #1450, but the idea of this PR is to make restarting tasks safer. We can have a situation where stale task workers (e.g. after a server restart; this happens a lot in dev) still exist and work is being done in parallel. If the stale task worker takes longer than the new one, it can overwrite the status etc. in the database. This PR prevents it from doing that, as each task worker gets the current execution generation of the task, and the generation counter is atomically incremented.

Ideally we'd just terminate the old worker processes, but after a restart we have no idea whether the PIDs are of our workers or not. They might have been reallocated. One solution I was thinking about was to have a communication protocol, so that they could be checked directly, but then workers would need some kind of concurrent interrupt processor, and I figured that would be more complicated than it's worth. We do still have pessimistic stale task detection, and I've been thinking about how to improve it.

### DIFF
--- a/atr/db/__init__.py
+++ b/atr/db/__init__.py
@@ -995,6 +995,7 @@ class Session(sqlalchemy.ext.asyncio.AsyncSession):
         added: Opt[datetime.datetime] = NOT_SET,
         started: Opt[datetime.datetime | None] = NOT_SET,
         pid: Opt[int | None] = NOT_SET,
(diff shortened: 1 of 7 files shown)


sbp:
I've decided to hold this feature as it is now less applicable to the later work than originally intended.

Issue #1532, opened 2026-08-26
Update KEYS file based on changes detected in SVN


alitheg:
ATR has 2 modes currently: manually update keys in ATR and sync to SVN, or automatically manage. There's a third required - don't manage in ATR but make sure ATR reflects changes made in SVN externally. That should use the existing pubsub watcher to detect changes and update KEYS in ATR, and should be a configurable mode on the committee page.

PR #1535, opened 2026-08-27
#1532 - add key sync from SVN

alitheg:
Add key sync from SVN and update the mode selection UI to include this. Because we decided keys shouldn't be removable once used, I also added a flag to keys removed from SVN but still in use.


sbp (changes requested):
Could this do a sync when (a) the reflect mode is first selected for older committees, or (b) when a new committee is detected in reflect mode? The first is less likely to be a problem because I only recently did the backfill, but I expect the latter might be confusing as the net effect is that nothing gets imported until the next `KEYS` change.

There are also a couple of "you can't edit this" methods on ATR that haven't been blocked in refusal mode: you can change which committees your _own_ keys are linked to, and you can also delete your own key.

Otherwise, looks good!

alitheg:
Great ideas and good catch on the refusals (or lack of). Will sort these and wrap it up.

PR #1561, opened 2026-09-02
#1554 - SVN rm on archive

### LINKED ISSUE #1554: Automatically purge old releases on svn dist/release
Currently only committing the new version to dist/release is automated. Would be beneficial to also automate removal of old versions (as described in <https://releases.apache.org/docs/promoting-to-release#removing-superseded-releases>). To detect old releases one can probably come up with a heuristics approach based on the newly released file names.

alitheg:
Add SVN rm command on all archive paths EXCEPT those triggered from the SVN dist watcher (since it's detecting an rm itself) and those in the catalog admin page (where you're manipulating data which didn't flow through ATR (from an import script or the dist watcher) and whose files will therefore be managed by the dist watcher. If we want to allow ATR to remove SVN files for non-ATR releases that's a different process!

### DIFF
--- a/atr/storage/writers/release.py
+++ b/atr/storage/writers/release.py
@@ -37,6 +37,7 @@ import sqlalchemy.engine as engine
 import sqlmodel
 
 import atr.analysis as analysis
+import atr.attestable as attestable
 import atr.auditlog as auditlog
 import atr.catalog_site as catalog_site
 import atr.config as config
@@ -49,7 +50,7 @@ import atr.hashes as hashes
 import atr.log as log
 import atr.models.api as api
 import atr.models.args as args
-import atr.models.attestable as attestable
+import atr.models.attestable as attestable_models
 import atr.models.results as results
 import atr.models.safe as safe
 import atr.models.sql as sql
@@ -107,17 +108,17 @@ class DeleteFileResult:
     metadata_files_deleted: int
 
 
-async def _archive_release(
+async def archive_release_core(
     data: db.Session,
     write_as: storage.WriteAs,
     asf_uid: str,
     project_key: safe.ProjectKey,
     version_key: safe.VersionKey,
     release: sql.Release,
+    source: sql.ArchiveSource,
 ) -> str | None:
-    # The archive core, shared by a committee member archiving by hand and the
-    # system archiving a dist removal. Only an announced release can be
-    # archived.
+    # Only an announced release can be archived. The source is recorded so the removal
+    # from SVN can be gated on it and a restore can refuse a release whose files have left dist.
     if release.phase != sql.ReleasePhase.RELEASE:
         return f"Release {project_key!s} {version_key!s} is not in the release phase"
 
@@ -127,13 +128,17 @@ async def _archive_release(
         sqlmodel.update(sql.Release)
         .where(via(sql.Release.key) == release.key)
         .where(via(sql.Release.is_archived).is_(False))
-        .values(archived=archive_date, is_archived=True)
+        .values(archived=archive_date, is_archived=True, archive_source=source)
     )
     update_result = await data.execute_query(update_stmt)
     if getattr(update_result, "rowcount", 0) != 1:
         return f"Release {project_key!s} {version_key!s} is already archived"
 
-    # TODO: SVN move to archive.apache.org goes here once SVN is wired up.
+    # The dist watcher archives a release because its files already left dist, so
+    # there's nothing left to remove - every other source is a live release we're
+    # taking out of the dist area.
+    if source is not sql.ArchiveSource.DIST_WATCHER:
+        queue_svn_unpublish(data, asf_uid, release)
 
     data.add(
         sql.LifecycleEvent(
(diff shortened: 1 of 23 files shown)


atr/storage/writers/release.py
```diff
@@ -1773,6 +1820,119 @@ async def finalise_published_release(self, task_args: args.ReleaseFinalise) -> r
...
+        ).demand(datatypes.FailedError("Archived release not found for removal"))
+        committee = release.committee
+        if committee is None:
+            raise datatypes.FailedError("Release has no committee - Invalid state")
+
+        # The removal undoes the publish: take out only the files ATR recorded for this
+        # release - the artifact, its signature, checksum and SBOM - grouped under the
+        # directory each lives in. A directory the release owns outright is taken whole,
+        # which clears its files, any subdirectories and the emptied directory in one
+        # action; the committee's shared dist root only ever gives up our own files.
```
sbp: This can fail if there are other files in there. Here's an example that I had an agent actually run with this PR code:

```
PUBLISHED to project/1.0.0 by svn import:
  CHANGES.txt
  README.md
  apache-project-1.0.0-src.tar.gz
  apache-project-1.0.0-src.tar.gz.asc
  apache-project-1.0.0-src.tar.gz.cdx.json
  apache-project-1.0.0-src.tar.gz.cdx.json.asc
  apache-project-1.0.0-src.tar.gz.cdx.json.sha512
  apache-project-1.0.0-src.tar.gz.sha256
  apache-project-1.0.0-src.tar.gz.sha512
RECORDED by announce in 1 Artifact row(s), download_path_suffix='project/1.0.0':
  apache-project-1.0.0-src.tar.gz
  apache-project-1.0.0-src.tar.gz.asc
  apache-project-1.0.0-src.tar.gz.cdx.json
  apache-project-1.0.0-src.tar.gz.sha512
TASK RESULT: Removed 4 file(s) for project-1.0.0 from SVN as r2
LEFT IN SVN under project/ after unpublish:
  1.0.0/
  1.0.0/CHANGES.txt
  1.0.0/README.md
  1.0.0/apache-project-1.0.0-src.tar.gz.cdx.json.asc
  1.0.0/apache-project-1.0.0-src.tar.gz.cdx.json.sha512
  1.0.0/apache-project-1.0.0-src.tar.gz.sha256
```

There's an existing test for this that should prevent it, but it happens to use exactly the subset of files that _doesn't_ cause this problem. Just adding something like a `README.md` file or a checksum for the SBOM JSON would break it.

atr/storage/writers/release.py
```diff
@@ -1598,8 +1639,14 @@ async def complete_archive(
                 # blocking the release
                 await self.__data.commit()
                 return None
-            return await _archive_release(
-                self.__data, self.__write_as, self.__asf_uid, project_key, version_key, release
+            return await archive_release_core(
+                self.__data,
+                self.__write_as,
+                self.__asf_uid,
```
sbp: During CAP resolve, this is `system`. Don't we always have to use a real ASF UID when committing to SVN?

atr/storage/writers/release.py
```diff
@@ -1773,6 +1820,119 @@ async def finalise_published_release(self, task_args: args.ReleaseFinalise) -> r
...
+        # already gone; any other error (a connection or auth failure) is no such proof,
+        # so it stays a failure.
+        rel_paths: list[str] = []
+        for directory, names in sorted(wanted.items()):
+            dir_url = f"{root_url}/{directory}"
+            if directory == committee_root:
+                # The shared dist root holds KEYS and other releases, so never take the
+                # directory itself, only our own files still in it. Non-recursive: a flat
+                # root may hold thousands of releases there's no reason to walk.
+                present = await self.__list_present(dir_url, recursive=False)
```
sbp: If a committee publishes a release straight into their committee directory, with no prefix, and there are subdirectories, `recursive=False` means that the listing call here won't find those files. A recursive list in SVN can still be done with a single `svn list` command, but it's true that the output might be huge.

sbp (changes requested):

atr/storage/writers/release.py
alitheg: Yes probably - do we have such a thing as an ATR service account we can use in that case?

atr/storage/writers/release.py
alitheg: Aha I didn't think of files that aren't artifacts. I'll use the attestable data's snapshot of the paths

atr/storage/writers/release.py
sbp: Nope, I think we'd have to request one for this.

dave2wave:
>  If we want to allow ATR to remove SVN files for non-ATR releases that's a different process!

We definitely need to do this to make transition to ATR easier.

atr/storage/writers/release.py
alitheg: I figured out we *can* just get the user who started the archive vote's uid and use that

sbp (commented):
Only thing I noticed is that there's a TOCTOU between listing and deleting. We can pass the revision in to the deletion so that we know it hasn't drifted since the listing.

Otherwise, all fine!

PR #1592, opened 2026-09-11
Fix release announcement recipients and sandboxed rsync uploads

tisonkun:
## Pull request summary

**Meaningful subject (required):** Fix release announcement recipients and sandboxed rsync uploads

**Description:**

Release announcements addressed to `announce@apache.org` pass the announcement form and storage validation, but the queued message task rejects them with `You are not permitted to send emails to announce@apache.org`. Allow that exact address in the worker's recipient validation, while preserving sender validation and restrictions on other root-domain accounts and external recipients. Cover permitted and rejected recipients in To, Cc, and Bcc, including rejection before the mail writer runs.

The Build and test workflow also exposed an independent rsync upload failure: the Alpine image now installs rsync 3.5.0, whose receiver opens each ancestor of an absolute destination path. Landlock denies access to those ancestors, so uploads exit with code 3. Start the upload subprocess in its quarantine directory and use exactly `.` as the receiver destination. This avoids reopening ancestors while preserving the existing sandbox grants, upload limits, and read behavior. The existing Playwright SSH upload test covers this regression.

Two worker-termination tests used a fake process group (`999`) without mocking `os.killpg`, so they could send real signals and fail with `PermissionError` depending on the runner. Mock signal delivery in both tests and assert the expected SIGTERM/SIGKILL sequence, preserving their deadline and surviving-child assertions.

## Validation

- The announcement regression failed before the recipient fix; all 26 message tests pass afterward.
- `uv run --frozen pytest -rs tests`: 2266 passed, 29 skipped.
- With a temporary fixture simulating permission-denied process-group signals, the termination tests reproduced the two CI failures before the test fix; all 10 pass afterward, on both macOS and Linux with the locked pytest and pytest-asyncio versions.
- `make check`: all checks passed.
- `docker build -t atr-pr1592-ci-fix -f Dockerfile.alpine .`: passed, with rsync 3.5.0 in the runtime image.
- Reproduced the rsync code 3 failure in the Linux runtime image with an absolute destination. The same sandboxed transfer passes with the destination set to `.` and the subprocess working directory set to the upload directory, including nested files.
- `playwright/test.py --skip-slow`: all tests passed against the rebuilt Linux application container, including the previously failing SSH upload. Verified with both the local macOS runner and a Linux runner using GNU rsync 3.2.7 (the same client version as CI). The Linux runner used cached Chromium with Playwright 1.61.0; the application image used rsync 3.5.0.
- All five existing sandbox tests passed in the Linux runtime image. A transfer through a symlink to a directory outside the sandbox was also rejected, with the outside file unchanged.
- The separate `tests/e2e/` suite was previously run against this branch: 109 passed, 7 skipped, 1 failed. Its sole failure, `test_announce_archives_prior_release`, also fails at the same archived-card assertion against unmodified `main` (`0ed47e0e`). This is separate from the Playwright workflow failure fixed here.
- All GitHub checks pass for commit `3285b72b`: [Build and test](https://github.com/apache/tooling-trusted-releases/actions/runs/34619580703), [pre-commit analysis](https://github.com/apache/tooling-trusted-releases/actions/runs/34619580709), and [CodeQL](https://github.com/apache/tooling-trusted-releases/actions/runs/34619580267). The hosted Playwright suite, including the SSH upload, completed successfully.

## Required acknowledgements

- [x] I have read and followed **CONTRIBUTING.md**
- [x] I have read **DEVELOPMENT.md**
- [x] I have run the required tests and checks locally
- [ ] All required checks are currently passing (GitHub CI passes; the separate pre-existing local e2e failure is documented above)
- [x] This branch is **rebased on the current `main` branch**

## Rebase confirmation details

The branch is based on `origin/main` at `0ed47e0e`. The rsync correction and worker-test isolation are separate follow-up commits.

### DIFF
--- a/atr/ssh.py
+++ b/atr/ssh.py
@@ -290,11 +290,12 @@ async def server_stop(server: asyncssh.SSHAcceptor) -> None:
     log.info("SSH server stopped")
 
 
-def _build_rsync_write_argv(argv: list[str], path: safe.StatePath) -> list[str]:
+def _build_rsync_write_argv(argv: list[str]) -> list[str]:
     """Build the rsync command for a write, adding enforced server side limits."""
     if len(argv) < 2 or argv[-2] != ".":
         raise RuntimeError("Validated rsync write argv must end with '.' and the destination path")
-    return [*argv[:-2], f"--max-size={_RSYNC_MAX_UPLOAD_SIZE}", "--info=skip2", ".", str(path)]
+    # Use exactly '.' so rsync does not reopen ancestors outside the Landlock sandbox
+    return [*argv[:-2], f"--max-size={_RSYNC_MAX_UPLOAD_SIZE}", "--info=skip2", ".", "."]
 
 
 async def _drain_stderr(stream: asyncio.StreamReader) -> bytes:
@@ -677,13 +678,13 @@ async def _step_07b_process_validated_rsync_write(
             nonlocal exit_status
             if old_rev is not None:
                 log.info(f"Using old revision {old_rev.number} and interim path {path}")
-            rsync_argv = _build_rsync_write_argv(argv, path)
+            rsync_argv = _build_rsync_write_argv(argv)
             rsync_argv = sandbox.command(rsync_argv, rw_paths=[str(path)])
 
             ###################################################
             ### Calls _step_08_execute_rsync_upload_command ###
             ###################################################
-            exit_status = await _step_08_execute_rsync(process, rsync_argv)
+            exit_status = await _step_08_execute_rsync(process, rsync_argv, cwd=path)
             if exit_status != 0:
                 if old_rev is not None:
                     for_revision = f"successor of revision {old_rev.number}"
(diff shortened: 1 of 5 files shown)


atr/ssh.py
```diff
@@ -290,11 +290,12 @@ async def server_stop(server: asyncssh.SSHAcceptor) -> None:
     log.info("SSH server stopped")
 
 
-def _build_rsync_write_argv(argv: list[str], path: safe.StatePath) -> list[str]:
+def _build_rsync_write_argv(argv: list[str]) -> list[str]:
     """Build the rsync command for a write, adding enforced server side limits."""
     if len(argv) < 2 or argv[-2] != ".":
         raise RuntimeError("Validated rsync write argv must end with '.' and the destination path")
-    return [*argv[:-2], f"--max-size={_RSYNC_MAX_UPLOAD_SIZE}", "--info=skip2", ".", str(path)]
+    # Use exactly '.' so rsync does not reopen ancestors outside the Landlock sandbox
```
dave2wave, line 297: It looks like an LLM told you this was a path traversal issue, we've seen that numerous times. And this was likely how we want it to work. @sbp ?

dave2wave: It would be best to make multiple PRs and issues. The fix in your initial PR is for a Blocker issue. Where as this one is for a false positive, and better raised as an Issue first.

dave2wave (dismissed):
Submit two PRs and not one.

sbp (approved):

sbp:
@tisonkun Thank you very much!

